PR #44, opened 2024-11-04
try/catch for redis connection

MoLala1:


src/main/routes/home.ts
```diff
@@ -3,10 +3,26 @@ import { Application } from 'express';
...
+    let logMsg = '';
+    if (!app.locals.redis) {
+      try {
+        app.locals.redis = new Redis(config.get('pcs.redis-connection-string'), {
+          maxRetriesPerRequest: 3,
+        });
+        await app.locals.redis.set('mykey', 'value');
+      } catch (error) {
+        logMsg = `${error.message} - Unable to connect and Set to redis`;
+        app.locals.redis.disconnect();
```
adusumillipraveen, line 16: A potential issue for future that app might still serve traffic without connection

adusumillipraveen (approved):

src/main/routes/home.ts
```diff
@@ -3,10 +3,26 @@ import { Application } from 'express';
...
 export default async function (app: Application): Promise<void> {
-  const redis = new Redis(config.get('pcs.redis-connection-string'));
-  await redis.set('mykey', 'value');
   app.get('/', async (req, res) => {
-    const mykey = await redis.get('mykey');
-    res.render('home', { redisKey: mykey });
+    let logMsg = '';
+    if (!app.locals.redis) {
+      try {
+        app.locals.redis = new Redis(config.get('pcs.redis-connection-string'), {
```
adusumillipraveen, line 10: May be the issue is we are trying connection on every request, try to move to a new method and create a new one if its not null. Something like 

```
let redisClient;

function getRedisClient() {
  if (!redisClient) {
    redisClient = new Redis(config.get('pcs.redis-connection-string'), {
      maxRetriesPerRequest: 3,
    });
    redisClient.on('error', (err) => {
      console.error(`Redis error: ${err.message}`);
    });
  }
  return redisClient;
}
```

src/main/routes/home.ts
MoLala1: i will add a ticket in the backlog to serve the traffic connectivity as means of keeping the design future

src/main/routes/home.ts
MoLala1: the code template will change as this would be an abstraction away from the route the validations but will create a new method and a validation for null checks.

PR #80, opened 2024-11-18
HDPI-82 - Add local redis cache

JakeCG:
### Jira link

[HDPI-82](https://tools.hmcts.net/jira/browse/HDPI-82)

### Change description

As an Infrastructure engineer

I want to add Redis to local setup for frontend

So that it aids devs in running application locally

### Testing done

Local redis cache added and auto pinged.

### Checklist

- [x] README and other documentation has been updated / added (if needed)

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -19,13 +19,13 @@ services:
   pcs-frontend-cache:
     image: bitnami/redis:7.4.1
     ports:
-      - "6379:6379"
+      - '6379:6379'
     volumes:
       - frontend-cache-data:/bitnami/redis/data
     environment:
       - ALLOW_EMPTY_PASSWORD=yes
     healthcheck:
-      test: [ "CMD", "redis-cli", "ping" ]
+      test: ['CMD', 'redis-cli', 'ping']
       interval: 3s
       timeout: 2s
       retries: 10
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,5 +1,5 @@
 services:
-  frontend:
+  pcs-frontend:
     build:
       context: .
       args:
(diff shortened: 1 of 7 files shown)


src/main/modules/session/index.ts
```diff
@@ -9,8 +9,12 @@ const { Logger } = require('@hmcts/nodejs-logging');
 export class Session {
   logger = Logger.getLogger('session');
   enableFor(app: Express): void {
-    const redis = new Redis(config.get<string>('secrets.pcs.redis-connection-string'));
+    const redisConnectionString =
+      process.env.REDIS_CONNECTION_STRING || config.get<string>('secrets.pcs.redis-connection-string');
+    this.logger.info(`Connecting to Redis at: ${redisConnectionString}`);
```
adusumillipraveen, line 14: Do not log the connection string please

docker-compose.yml
```diff
@@ -10,5 +7,34 @@ services:
...
+    depends_on:
+      pcs-frontend-cache:
+        condition: service_healthy
+    environment:
+      - REDIS_CONNECTION_STRING=redis://pcs-frontend-cache:6379
+    networks:
+      - frontend-network
+
+  pcs-frontend-cache:
+    image: bitnami/redis:7.4.1
```
adusumillipraveen, line 20: We are on Azure Redis version 6, but i don't think it might cause any issues. @MoLala1 ?

MoLala1: it technically shouldnt be an issue as in charts.yaml: https://github.com/hmcts/pcs-frontend/blob/1cf213dd7bf0529f7522fd62a606353cf6ea9615/charts/pcs-frontend/Chart.yaml#L11

am using image version 20.3.0 but please test to be sure Jake :-)

MoLala1 (approved):
approved

docker-compose.yml
adusumillipraveen: It's not an issue currently, but we are developing against a latest version in local and someone might make a change and it to realise it won't work in Azure as azure instances are o old version. But happy to try out and change if we see such issues in future.

src/main/modules/session/index.ts
MoLala1: was only for testing as part of my initial implementation, which i clearly didnt remove :-)

src/main/modules/session/index.ts
```diff
@@ -9,7 +9,9 @@ const { Logger } = require('@hmcts/nodejs-logging');
 export class Session {
   logger = Logger.getLogger('session');
   enableFor(app: Express): void {
-    const redis = new Redis(config.get<string>('secrets.pcs.redis-connection-string'));
+    const redisConnectionString =
+      process.env.REDIS_CONNECTION_STRING || config.get<string>('secrets.pcs.redis-connection-string');
```
adusumillipraveen, line 13: this should work as there is env var mapping already?

```suggestion
      config.get<string>('secrets.pcs.redis-connection-string');
```

adusumillipraveen (approved):

PR #149, opened 2024-12-11
added S2S

MoLala1:
https://tools.hmcts.net/jira/browse/HDPI-86


src/main/modules/s2s/index.ts
```diff
@@ -0,0 +1,71 @@
...
+
+        try {
+          req.session.serviceToken = (await request) as string;
+        } catch (error) {
+          this.logger.error('S2S ERROR', error.message);
+        }
+      }
+
+      if (req.session.serviceToken) {
+        this.logger.info('SERVICE TOKEN = ', s2sUrl, req.session.serviceToken);
```
JakeCG, line 65: Do we want to be logging the secret every time, even on success, won't this this drive up cost

src/main/modules/s2s/index.ts
```diff
@@ -0,0 +1,71 @@
...
+        const request = new Promise((resolve, reject) => {
+          fetch(`${s2sUrl}/lease`, {
+            method: 'POST',
+            headers: {
+              'Content-Type': 'application/json',
+            },
+            body: JSON.stringify(params),
+          })
+            .then(response => {
+              this.logger.info('S2S response code: ', response.status);
```
JakeCG, line 39: Same as the previous logging comment

src/main/routes/home.ts
```diff
@@ -1,13 +1,18 @@
...
-  app.get('/', (req: Request, res: Response) => {
-    res.render('home');
+  app.get('/', async (req: Request, res: Response) => {
+    let apiResponse: Partial<AxiosResponse> = {
+      data: 'default value',
+    };
+    try {
+      apiResponse = await axios.get(config.get('api.url'));
+    } catch (error) {
+      // console.error('pcs-api error', error.response.statusText);
```
JakeCG, line 13: This probably can be removed / replaced

src/main/modules/s2s/index.ts
adusumillipraveen: Not just cost, but its sensitivie enough to be in logs.

JakeCG (approved):

PR #315, opened 2025-04-16
HDPI-611: Refactor out PCS API client

scottstewart-sl:
### Jira link

See [HDPI-611](https://tools.hmcts.net/jira/browse/HDPI-611)

### Change description

Encapsulate the code that calls the PCS API client into a single shared module.

### Testing done

There are no functional changes as part of this refactoring. I have run the unit tests locally.

### Checklist


src/main/modules/pcs-api-client/index.ts
paddy-hmcts, line 1: Not sure this belongs as a module, as we have to 'new' up an instance when required? Could definitely have a service class for this

src/main/modules/pcs-api-client/index.ts
```diff
@@ -0,0 +1,22 @@
...
+import { CourtVenue } from './court-venue.interface';
+
+export class PcsApiClient {
+  private readonly pcsApiURL: string;
+
+  constructor() {
+    this.pcsApiURL = config.get<string>('api.url');
+  }
+
+  async getRootGreeting(): Promise<string> {
```
paddy-hmcts, line 13: wondering if this is still required? initially was more POC to test integration with the api, which has now bene proven

scottstewart-sl: As discussed I will leave it in here until we have another API call in place in the `master` branch, then it can be removed.

src/main/modules/pcs-api-client/index.ts
scottstewart-sl: Yes, good suggestion. I have now refactored it as we discussed and it looks cleaner.

MoLala1 (approved):

PR #331, opened 2025-04-24
Hdp1 347 idam backend

paddy-hmcts:
### Jira link
https://tools.hmcts.net/jira/browse/HDPI-347

### changes

Fixes OIDC development locally (node + docker)
Fixes https locally (node + docker)

Fixes issue with shutdown


.env.example
```diff
@@ -0,0 +1,2 @@
+ OIDC_CLIENT_SECRET=AAA
+ S2S_SECRET=AAA
```
scottstewart-sl, line 2: If the intention is for the developer to copy this to `.env` locally then could you add a note to the README? (Assuming there is no other process that is already doing it automatically).

elvianihuseyin2 (approved):

MoLala1 (approved):

src/main/server.ts
```diff
@@ -40,11 +44,25 @@ function gracefulShutdownHandler(signal: string) {
...
-    // Close server if it's running
-    httpsServer?.close(() => {
-      logger.info('HTTPS server closed');
+    app.locals.redisClient?.quit(async (err: Error | null) => {
+      if (err) {
+        logger.error('Error quitting Redis client', err);
+      }
+      logger.info('Redis client quit');
+
+      if (app.locals.ENV === 'development') {
```
scottstewart-sl, line 53: Minor - it might be cleaner to have two separate top level variables for `httpsServer` and `httpServer`, (they are different types anyway), and then just call the shutdown steps here for the one that is defined, rather than having to match the conditional logic at the top that was used when starting one.

In other words, at this point in the code it shouldn't matter what condition originally caused an `httpsServer` or an `httpServer` to be created, just the fact that whichever one was created needs to be shutdown.

.env.example
MoLala1: good shout scott. will add it in.

src/main/server.ts
paddy-hmcts: I did try this, however it depends on the type created on how it shuts down

PR #334, opened 2025-04-25
Check node version before running application in Dev

ed-graham:
### Jira link

See [PROJ-XXXXXX](https://tools.hmcts.net/jira/browse/PROJ-XXXXXX)

### Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,11 +3,11 @@
   "version": "0.0.1",
   "private": true,
   "engines": {
-    "node": ">=18.0.0"
+    "node": ">=22.0.0"
   },
   "scripts": {
     "start:dev:ssl": "./bin/generate-ssl-options.sh && NODE_ENV=development OIDC_REDIRECT_URI=https://localhost:3209/oauth2/callback ts-node -r tsconfig-paths/register --files src/main/server.ts",
-    "start:dev": "NODE_ENV=dev ts-node -r tsconfig-paths/register --files src/main/server.ts",
+    "start:dev": "node -v | grep -Eq 'v2[2-9]|v[3-9][0-9]|v[1-9][0-9]{2,}' && (NODE_ENV=dev ts-node -r tsconfig-paths/register --files src/main/server.ts || exit 0) || (echo Incompatible version of Node && node -v && exit 1)",
     "start": "NODE_ENV=production dist/main/server.js",
     "start:watch": "nodemon --config nodemon.json",
     "docker:dev": "docker compose up",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -3,11 +3,11 @@
   "version": "0.0.1",
   "private": true,
   "engines": {
-    "node": ">=18.0.0"
+    "node": ">=22.0.0"
   },
   "scripts": {
     "start:dev:ssl": "./bin/generate-ssl-options.sh && NODE_ENV=development OIDC_REDIRECT_URI=https://localhost:3209/oauth2/callback ts-node -r tsconfig-paths/register --files src/main/server.ts",
-    "start:dev": "NODE_ENV=dev ts-node -r tsconfig-paths/register --files src/main/server.ts",
+    "start:dev": "node -v | grep -Eq 'v2[2-9]|v[3-9][0-9]|v[1-9][0-9]{2,}' && (NODE_ENV=dev ts-node -r tsconfig-paths/register --files src/main/server.ts || exit 0) || (echo Incompatible version of Node && node -v && exit 1)",
```
paddy-hmcts, line 10: could this line be reverted please, the package manager should handle incompatible versions based on the engines above

paddy-hmcts (changes requested):

package.json
ed-graham: The package manager doesn't handle incompatible versions based on the engines.  I tried that first of all; but when it had no effect, I then added this line.  I've been running `yarn start:dev` with Node v18 and the package manager lets me do it; hence the need for this change!

ed-graham: I created an `.npmrc` file and added the line `engine-strict=true` but it had no effect.  I also tried creating a `.yarnrc` file (since we are using yarn) but that had no effect either.

ed-graham: I would love this to work just by using the engines!  But I didn't find a way of doing it just with them.

MoLala1: i would disagree, we shouldn't manually have to be reliant on the validations of versioning.  Whilst i understand it hasnt worked for you specifically but its a working pattern.

ed-graham: I'm not sure I understand what you mean.  At the moment, if you run `yarn start:dev` with Node v18 you get a run-time error with a stack dump that suggests the code is incorrect when actually it's fine, it's the version of Node that's wrong.  That can't be a good result, surely?!

paddy-hmcts: https://github.com/hmcts/pcs-frontend/blob/master/README.md if you look at prequisites it specifically states v22, we shouldn't have to worry at this point

ed-graham: @paddy-hmcts - but that's precisely my point.  We do still have to worry.  I had installed v22 and yet I still had this problem, so others coming after me could have it as well.

elvianihuseyin2: Ehm i think it is a bit overkill to check node version every single time we do yarn start

PR #337, opened 2025-04-29
Hdpi 500 dashboard notificatons UI

MoLala1:
### Jira link
https://tools.hmcts.net/jira/browse/HDPI-500


src/main/views/components/notification/Notice.AAA3.ClaimIssue.ClaimSubmit.Required.njk
```diff
@@ -0,0 +1,12 @@
+<div class="govuk-notification-banner" role="region" aria-labelledby="govuk-notification-banner-title" data-module="govuk-notification-banner">
+  <div class="govuk-notification-banner__header">
+    <h2 class="govuk-notification-banner__title" id="govuk-notification-banner-title">
+      Trial or hearing scheduled
+    </h2>
+  </div>
+  <div class="govuk-notification-banner__content">
+    Your appointment is on {{ params.dueDate | date('d LLLL y \'at\' h:mma') }} in {{ params.location }}. You need to
```
scottstewart-sl, line 8: This first placeholder should be using `params.appointmentTime`

src/main/routes/dashboard.ts
```diff
@@ -0,0 +1,21 @@
...
+  app.get('/dashboard/:caseReference', oidcMiddleware, async (req: Request, res: Response) => {
+    const caseReference: number = parseInt(req.params.caseReference, 10);
+    try {
+      const notifications: DashboardNotification[] = await getDashboardNotifications(caseReference);
+      res.render('dashboard', {
+        notifications,
+      });
+    } catch (e) {
+      logger.error(`Failed to fetch notifications for case ${caseReference}. Error was: ${e}`);
+      throw e;
```
scottstewart-sl, line 18: I think the agreement, (on the Jira), was that an error fetching the notifications would show a notification with an error message, at least until we get requirements from UCD,

src/test/unit/modules/nunjucks/filters/date.test.ts
```diff
@@ -0,0 +1,43 @@
...
+
+  it('should format date with custom format', () => {
+    const isoDate = '2024-03-20T12:00:00Z';
+    const customFormat = 'dd/MM/yyyy';
+    const result = date(isoDate, customFormat);
+    expect(result).toBe('20/03/2024');
+  });
+
+  it('should handle different timezones correctly', () => {
+    const isoDate = '2024-03-20T00:00:00Z';
```
scottstewart-sl, line 18: On 20/03/2024 UK time was the same as UTC so the time offset would have been the same. It's clearer to use a date in the middle of summer to be sure that it is BST. I usually go for 30th June.

Also it would be best to set the time as 23:00:00 Z so that it proves that the hour has been added to convert to UK time, (BST)

src/test/unit/modules/nunjucks/filters/currency.test.ts
```diff
@@ -0,0 +1,23 @@
...
+  it('should format whole numbers with two decimal places', () => {
+    expect(currency(100)).toBe('£100.00');
+  });
+
+  it('should format decimal numbers with two decimal places', () => {
+    expect(currency(99.99)).toBe('£99.99');
+  });
+
+  it('should format numbers with more than two decimal places by rounding', () => {
+    expect(currency(50.999)).toBe('£51.00');
```
scottstewart-sl, line 13: That's worth double checking with the BA / Product Owner. A currency value with more than 2 d.p. coming from the  API might suggest a data/code issue and silently rounding it might not be the best option in that case.

src/main/routes/dashboard.ts
MoLala1: showing no notifications to the user and logging a error in the logs is more relevant given the state of maturity of the designs which is why i went with the approach.

src/test/unit/modules/nunjucks/filters/currency.test.ts
MoLala1: valid point Scott.  Wondering whether at this point it is worth doing again with designs being fluid.

src/main/views/components/notification/Notice.AAA3.ClaimIssue.ClaimSubmit.Required.njk
MoLala1: actioned, thanks.

src/test/unit/modules/nunjucks/filters/currency.test.ts
MoLala1: its an edge case, granted however can occur never the less but yup will check with BA.

src/main/routes/dashboard.ts
scottstewart-sl: OK, in which case could you update the Jira ticket as that is what Dev Assurance and QA will be using to test the story?

Also, if I'm reading the code correctly, an error fetching the dashboard notifications from the API would cause the whole dashboard page to fail. Perhaps it would be better to catch and log the error from the API call and then render the dashboard with an empty list of dashboard notifications? (Particularly once other components like the task groups are added to the dashboard page which might not have error fetching data from the API).

MoLala1: will add and update the story ac with the ammendment. cheers

src/test/unit/modules/nunjucks/filters/currency.test.ts
MoLala1: spoken with BA(Nick) and he is content with the current approach implemented.

src/main/routes/dashboard.ts
MoLala1: notifications shouldn't form a critical part of the app and as discussed, will update once the final designs are in.

paddy-hmcts (approved):

elvianihuseyin2 (approved):

PR #346, opened 2025-05-01
Hdpi-645 create idam user

Gautham059:
https://tools.hmcts.net/jira/browse/HDPI-645

Bash script to get an access token and create an IDAM user.
This is the first stage to provide a method of creating users for manual testing. Subsequent work will be built on this to allow automated creation of users in Typescript.


README.md
```diff
@@ -102,6 +102,21 @@ the following command:
 yarn test
 ```
 
+Use this command to create a temporary IDAM user:
+
+```bash
+yarn createIdamUser <role> <email>
```
scottstewart-sl, line 108: Up to you, but you could list the other optional parameters here as well ([password] [surname] [forename]).

README.md
```diff
@@ -102,6 +102,21 @@ the following command:
...
+```
+
+- <role> space delimited list of user roles
+- <email> should not match any existing user email IDs
+
+For example, if you want to create an IDAM user with the email 'test@test.com' and the role 'citizen', use the following command:
+
+```bash
+yarn createIdamUser citizen test@test.com
```
scottstewart-sl, line 117: Could you update the example to show more than 1 role? Presumably the role list needs to be quoted?

README.md
Gautham059: Done

README.md
Gautham059: Done

seymour-quality (commented):
lgtm

scripts/get_secrets.sh
```diff
@@ -0,0 +1,40 @@
+#!/bin/bash
+# Change KEY_VAULT_NAME if you want to use a different environment
+# Secrets must be tagged with "test": "ENV_VAR_NAME" in Azure Key Vault
+# This script looks for the tagged secrets and creates an .env file based on the .env.example file
```
adusumillipraveen, line 4: Do we need this complexity ? We just need 1 secret which can directly go into the user creation script. This can also lead to .env file being left with secrets which people can accidentally commit to the repo.

seymour-quality (approved):

scripts/createIdamUser.sh
```diff
@@ -0,0 +1,62 @@
+#!/bin/bash
+## Usage: ./createIdamUser.sh roles email [password] [surname] [forename]
```
adusumillipraveen, line 2: Its better to use  args like `--password` in the script when we have optional. For example if someone wants to pass forename alone in this example, then they have to pass password and surname as well.

scripts/createIdamUser.sh
```diff
@@ -0,0 +1,62 @@
...
+           --data-urlencode 'grant_type=client_credentials' \
+           --data-urlencode 'client_id=pcs-frontend' \
+           --data-urlencode 'client_secret='${PCS_FRONTEND_IDAM_SECRET}'' \
+           --data-urlencode 'scope=profile roles')
+response=${response#*:\"}
+accessToken=${response%%\"*}
+
+rolesStr=$1
+email=$2
+password=${4:-Pa$$w0rd}
```
adusumillipraveen, line 26: Is ${3 missing ?

scripts/createIdamUser.sh
```diff
@@ -0,0 +1,62 @@
...
+export $(grep -v '^#' .env | xargs)
+
+response=$(curl --location 'https://idam-api.aat.platform.hmcts.net/o/token' \
+           --header 'Content-Type: application/x-www-form-urlencoded' \
+           --data-urlencode 'grant_type=client_credentials' \
+           --data-urlencode 'client_id=pcs-frontend' \
+           --data-urlencode 'client_secret='${PCS_FRONTEND_IDAM_SECRET}'' \
+           --data-urlencode 'scope=profile roles')
+response=${response#*:\"}
+accessToken=${response%%\"*}
```
adusumillipraveen, line 22: Use Jq, something like 

accessToken=$(echo "$response" | jq -r .access_token)
```suggestion
accessToken=${response%%\"*}
```

Jenkinsfile_CNP
```diff
@@ -10,7 +10,23 @@ def component = "frontend"
 
 def yarnBuilder = new uk.gov.hmcts.contino.YarnBuilder(this)
 
+static Map < String, Object > secret(String secretName, String envVariable) {
```
adusumillipraveen, line 13: This is not needed for this PR, but no harm as this repo might need it in future for tests.

scripts/get_secrets.sh
Gautham059: .env is in .gitignore already(Like for other repos) and shouldn't we be worried about committing it to this repo? And we still might be needing this file for local testing any way?

adusumillipraveen: Even leaving them on someone's machine makes them vulnerable to be exposed. They should be just exposed as env variables which wipeout with the lifecycle of process/shell.

Jenkinsfile_CNP
Gautham059: Yes, this will facilitate downstream work which is being developed now.

scripts/createIdamUser.sh
Gautham059: Updated the script to support command line arguments

scripts/createIdamUser.sh
Gautham059: Resolved this.

scripts/createIdamUser.sh
Gautham059: Updated the script

PR #364, opened 2025-05-13
feat: Hdpi 511 dashboard task groups

paddy-hmcts:
### Jira link

https://tools.hmcts.net/jira/browse/HDPI-511


src/main/assets/fixtures/taskgroups.json
```diff
@@ -0,0 +1,116 @@
...
+        "templateValues": {},
+        "status": "AVAILABLE"
+      },
+      {
+        "templateId": "Task.AAA6.Hearing.UploadDocuments",
+        "templateValues": {
+          "deadline": "2025-05-20T15:00:00.000Z"
+        },
+        "status": "ACTION_NEEDED"
+      },
```
scottstewart-sl, line 56: This looks like test data. Shouldn't it be under the test folder?

paddy-hmcts: it's to stub the data from pcs-api which is incorrectly done, once the api is ready this will all be removed anyway

src/main/routes/dashboard.ts
```diff
@@ -1,20 +1,94 @@
...
+}
+
+const mapTaskGroups =
+  (app: Application, caseReference: number) =>
+  (taskGroups: DashboardTaskGroup[]): MappedTaskGroup[] => {
+    return taskGroups.map(taskGroup => {
+      const mappedTitle = TASK_GROUP_MAP[taskGroup.groupId];
+      return {
+        groupId: taskGroup.groupId,
+        title: mappedTitle,
```
scottstewart-sl, line 36: What will happen if there isn't an entry in `TASK_GROUP_MAP` for `taskGroup.groupId`? Is it going to show "null" in the UI?

I think it would be better to handle this explicitly in a suitable way, (that can be decided by the business), ranging from erroring the whole page to just logging the error and leaving that task out of the rendered content.

paddy-hmcts: agreed, and yes I didn't test that. The ticket didn't specify any unhappy paths tbh, but yes we need to define an overall process for error handling

src/main/routes/dashboard.ts
```diff
@@ -1,20 +1,94 @@
...
+
+          return {
+            title: {
+              html: app.locals.nunjucksEnv.render(
+                `components/taskGroup/${taskGroupId}/${task.templateId}.njk`,
+                task.templateValues
+              ),
+            },
+            hint,
+            // TODO: slugify the templateId
```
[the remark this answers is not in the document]
paddy-hmcts: no, as those pages/urls won't exist at moment anyway

src/main/routes/dashboard.ts
paddy-hmcts: image
So I did investigate this, currently it errors the whole page

MoLala1 (approved):

PR #372, opened 2025-05-15
fix: s2s token not being picked up correctly, refactored

paddy-hmcts:
### Jira link
https://tools.hmcts.net/jira/browse/HDPI-511

### DIFF
--- a/config/custom-environment-variables.json
+++ b/config/custom-environment-variables.json
@@ -18,7 +18,9 @@
     "ttlInSeconds": "REDIS_TTL"
   },
   "s2s": {
-    "url": "S2S_URL"
+    "url": "S2S_URL",
+    "ttl": "S2S_TTL",
+    "key": "S2S_KEY"
   },
   "oidc": {
     "issuer": "OIDC_ISSUER",
--- a/src/main/modules/http/index.ts
+++ b/src/main/modules/http/index.ts
@@ -79,7 +79,7 @@ export class HttpService {
           // Clear the token if regeneration failed
           this.s2sToken = null;
           this.tokenExpiry = null;
-          this.logger.error('[HttpService] Token regeneration error:', error);
+          this.logger.error('Token regeneration error:', error);
           throw error;
         }
       })().finally(() => {
--- a/src/main/modules/http/index.ts
+++ b/src/main/modules/http/index.ts
@@ -1,41 +1,93 @@
-import axios, { type AxiosInstance, type AxiosRequestConfig, type AxiosResponse } from 'axios';
-
+import { Logger } from '@hmcts/nodejs-logging';
+import axios, {
+  type AxiosInstance,
+  type AxiosRequestConfig,
+  type AxiosResponse,
+  type InternalAxiosRequestConfig,
+} from 'axios';
 type TokenRegenerator = () => Promise<void>;
 
-class HttpService {
-  private axiosInstance: AxiosInstance;
+export class HttpService {
+  private instance: AxiosInstance;
   private s2sToken: string | null = null;
   private tokenExpiry: number | null = null;
   private tokenRegenerator: TokenRegenerator | null = null;
-  private isRegenerating: boolean = false;
+  private tokenRegenerationPromise: Promise<void> | null = null;
+  private readonly TOKEN_WAIT_TIMEOUT = 5000; // 5 seconds
+  private readonly TOKEN_WAIT_INTERVAL = 100; // 100ms
+  logger = Logger.getLogger('http');
 
   constructor() {
-    // Create axios instance with interceptor
-    this.axiosInstance = axios.create();
-    this.axiosInstance.interceptors.request.use(async config => {
+    this.instance = axios.create();
+
+    // Request interceptor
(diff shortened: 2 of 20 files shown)


src/main/modules/s2s/index.ts
```diff
@@ -4,59 +4,78 @@ import config from 'config';
...
+              .catch((err: string) => {
+                this.logger.error(err);
+                reject(err);
+              });
+          });
+
+          try {
+            serviceToken = await request;
+            // Store token in Redis with expiry so it is automatically removed after TTL
+            await redisClient?.set(s2sKey, serviceToken, 'EX', s2sTtl);
```
scottstewart-sl, line 61: How about splitting these two statements out into separate try/catch blocks? A failure to store the token in Redis will just result in a slight result in efficiency, but the app will carry on working. However, a failure to get a new s2s token is more significant and in my opinion it would be better to fail fast here with a specific log error message.

src/main/modules/s2s/index.ts
```diff
@@ -4,59 +4,78 @@ import config from 'config';
...
+          } catch (error) {
+            this.logger.error('S2S ERROR', error.message);
+          }
+        }
+
+        const currentHeader = axios.defaults.headers.common['ServiceAuthorization'];
+
+        if (serviceToken && currentHeader !== `Bearer ${serviceToken}`) {
+          // Only set the header if it's not already set with the same token
+          axios.defaults.headers.common['ServiceAuthorization'] = `Bearer ${serviceToken}`;
```
scottstewart-sl, line 71: I know I've raised this before, but I still think it is a security risk to set this header on _all_ Axios requests. It only takes a new developer to come along in a few months and implement an integration with a third party API and this s2s token will be exposed. Is there nowhere else it can be stored, (e.g. `app.locals`), and then just applied when calling internal APIs?

config/custom-environment-variables.json
```diff
@@ -15,7 +15,9 @@
     "ttlInSeconds": "REDIS_TTL"
   },
   "s2s": {
-    "url": "S2S_URL"
+    "url": "S2S_URL",
+    "ttl": "S2S_TTL",
```
scottstewart-sl, line 19: I would put the unit of time in this variable name everywhere it is referenced, (like `ttlInSeconds` above). It just avoids ambiguity

scottstewart-sl (commented):
A few small comments from me.

scottstewart-sl:
I've been thinking about this approach a bit more. I know the middleware approach is common in an app like this, but I wonder if it is the best thing to do for the requirement.

Firstly, based on my limited understanding, an s2s token is being fetched, (even if it is just from Redis), for every request even if the request won't need to call an internal API.

Secondly, if the token is invalid for some reason at the point when it is used then it should be renewed. That won't happen at the moment from what I can see, and the situation wouldn't resolve itself until the item in the Redis cache expires, (which looks like it has a TTL of 1 hour).

An alternative approach might be to have some sort of generic wrapper function which is basically, "call an internal API with an s2s token". The implementation would be something along the lines of:

- The API request details are passed to the function without a service auth header
- The function checks if `app.locals` has an s2s token and if not generates a new one and stores it back there
- The function sets the service auth header and makes the call with it.
- If the response is a 401 status then the function renews the s2s token, stores it, and tries once more. Other errors are propagated immediately.
- A second 401 response would be propagated

I think I've seen that pattern used in previous projects I have been involved in and it helps prevent getting stuck with in invalid cached auth token. I also don't really see the benefit of having the s2s tokens in Redis if there are only a small number of instance of the service running.

We can discuss further in a call if you want.

src/main/modules/http/index.ts
```diff
@@ -0,0 +1,77 @@
...
+    this.axiosInstance = axios.create();
+    this.axiosInstance.interceptors.request.use(async config => {
+      if (!this.s2sToken || this.isTokenExpired()) {
+        if (!this.tokenRegenerator) {
+          throw new Error('No valid S2S token available and no regenerator configured');
+        }
+
+        // Prevent multiple simultaneous regeneration attempts
+        if (this.isRegenerating) {
+          throw new Error('Token regeneration in progress');
```
scottstewart-sl, line 23: Might this approach generate an error if there are multiple concurrent requests to the app? I know that Node isn't multi-threaded but perhaps things are slightly more complicated with Axios promises

Would an approach along these lines be more robust? It looks like it returns the Promise from the in-progress token refresh and then any request can wait for it, rather than having to throw an Error:

https://github.com/axios/axios/issues/450#issuecomment-247446276

I don't understand the details of that code in the link, but maybe it just hooks into the error handler and would make single attempt at a token renewal for a 401 status and then somehow gets Axios to automatically retry the request? But even if it doesn't do the retry, using a shared Promise to wait on would hopefully avoid the need for that Error.

paddy-hmcts: This was a good suggestion and I've implemented the details in the code

src/main/modules/s2s/index.ts
paddy-hmcts: updated to use a 'http' service instead

MoLala1 (approved):

PR #416, opened 2025-06-09
DB Utility helper is added

madhavi1203:
HDPI-721 : Create DB Utility Module perform Database Operations(Connect, Query)

Description:

DB connection handling
Query execution with parameterisation
Error handling and logging
Configured to connect diff DB (AAT, Preview) and tested.

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -12097,8 +12097,8 @@ __metadata:
   linkType: hard
 
 "sass@npm:^1.65.1":
-  version: 1.89.1
-  resolution: "sass@npm:1.89.1"
+  version: 1.89.2
+  resolution: "sass@npm:1.89.2"
   dependencies:
     "@parcel/watcher": "npm:^2.4.1"
     chokidar: "npm:^4.0.0"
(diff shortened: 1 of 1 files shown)


paddy-hmcts (changes requested):

adusumillipraveen (commented):
I haven't had a close look at code, but I don't think frontend repos should be talking to database. A clear violation of layering and adding tight coupling. Has this been tested in preview? Our current DB in preview is not accessible from Jenkins, so i am not sure  how it even worked?

paddy-hmcts:
> I haven't had a close look at code, but I don't think frontend repos should be talking to database. A clear violation of layering and adding tight coupling. Has this been tested in preview? Our current DB in preview is not accessible from Jenkins, so i am not sure how it even worked?

I concur. I'm also wondering if it would be worth splitting up the frontend repo from the e2e tests (as these have a lot of dependencies)

src/test/test-config.ts
Gautham059, line 1: Can we remove this config file and use the default.json under root config folder?

PR #1056, opened 2026-04-10
Professionalrespondtoclaim

m014389b:
### Jira link

N/A

### Change description

Refactor to allow for legal representative user journey to reuse screens templates and contents where appropriate but also allow for json asset keys to be merged allowing for overriding.

Whether the legal representative journey should be loaded or not is determined by the request session user roles. This will then determine multiple steps: 
- whether to read additional json assets from "respondToClaim/legalrep/xxx"
- whether to load the legalrep flowconfig which will overload the step order and steps to allow for a different journey
- whether to load xui header and footers configuration is loaded (from hmcts-cft/cft-ui-component-lib)
- whether the XUI headers / footers are displayed (stepsTemplate.njk) 
- the contents of start now (will be expanded as more screens are migrated/ adapted) to legal representative journey

### Testing done

Unit test coverage

  * [ ] Yes
  * [x] No

### Checklist


src/main/middleware/legalRepresentativeHeaders.ts
```diff
@@ -0,0 +1,37 @@
...
+  req: Request,
+  res: Response,
+  next: NextFunction
+): Promise<void> => {
+  const isLegalRepresentative = isLegalRepresentativeUser(req);
+  let headerModel, footerModel;
+
+  if (isLegalRepresentative) {
+    const roles = req.session?.user?.roles;
+    const xuiBaseUri: string = process.env.XUI_BASE_URI || config.get('xui.uri');
```
scottstewart-sl: Is the env var check needed here if the mapping has already been added to `custom-environment-variables.json` ?

src/main/assets/locales/cy/respondToClaim/legalrep/defendantNameConfirmation.json
```diff
@@ -0,0 +1,9 @@
+{
+  "pageTitle": "cyYour client's name",
```
scottstewart-sl: Some of these apostrophes will need to change to the curly ones, (like the one you have in `you’ve submitted` in startNow.json).

src/main/steps/index.ts
```diff
@@ -1,38 +1,99 @@
...
   },
 };
 
+function getJourneyConfigForRequest(journeyName: string, req?: Request): ResolvedJourneyConfig | undefined {
+  const journey = journeyRegistry[journeyName];
+  if (!journey) {
+    return undefined;
+  }
+
+  const userType = req ? getUserType(req) : 'citizen';
```
scottstewart-sl, line 57: This `getUserType` would always return something wouldn't it? i.e. the default is encapsulated in that function, so doesn't need to be repeated here.

src/main/steps/index.ts
```diff
@@ -1,38 +1,99 @@
...
+      flowConfig: makeAnApplicationFlowConfig,
+      stepRegistry: makeAnApplicationStepRegistry,
+    },
   },
 };
 
+function getJourneyConfigForRequest(journeyName: string, req?: Request): ResolvedJourneyConfig | undefined {
+  const journey = journeyRegistry[journeyName];
+  if (!journey) {
+    return undefined;
```
scottstewart-sl, line 54: Is this effectively an error scenario from misconfiguration? If so then perhaps add a warning log here, so we know if it is happening.

src/main/assets/locales/cy/respondToClaim/legalrep/defendantNameConfirmation.json
m014389b: Done

src/main/middleware/legalRepresentativeHeaders.ts
m014389b: Good spot

src/main/steps/index.ts
m014389b: This check is for defence of "req" being undefined, in which case default to the "citizen". This occurs on LOC 129 of "registerSteps.ts"

src/main/steps/index.ts
m014389b: Yes, we _shouldn't_ get into this scenario. Logging makes sense

src/main/steps/index.ts
scottstewart-sl: Sorry, I misread the code. 🙂

scottstewart-sl (approved):

src/main/middleware/legalRepresentativeAccess.ts
```diff
@@ -0,0 +1,26 @@
...
+    return next();
+  }
+
+  const isAllowed = LEGAL_REPRESENTATIVE_ALLOWED_PATHS.some(pattern => pattern.test(req.path));
+
+  if (isAllowed) {
+    return next();
+  }
+
+  res.status(404).send('Not Found');
```
paddy-hmcts, line 25: I think as this is restricting access to the user shouldn't it be a 403?

src/main/middleware/legalRepresentativeAccess.ts
```diff
@@ -0,0 +1,26 @@
+import type { NextFunction, Request, RequestHandler, Response } from 'express';
+
+import { isLegalRepresentativeUser } from '../steps/utils';
+
+const LEGAL_REPRESENTATIVE_ALLOWED_PATHS = [
+  /^\/case\/[^/]+\/respond-to-claim(?:\/.*)?$/,
+  /^\/api\/postcode-lookup(?:\/.*)?$/,
+];
```
paddy-hmcts, line 8: not overly keen on relying on regex or paths, think this should just be a middleware that purely checks the roles, and then it should be applied on a per route/routerr basis?

src/main/middleware/legalRepresentativeAccess.ts
scottstewart-sl: I think it depends. Sometimes a 404 is used to hide the fact that a page exists - for example if a user tries to respond to a claim for a case reference that they aren't linked to, then they will get a 404, but the case does exist.

src/main/middleware/legalRepresentativeAccess.ts
m014389b: The idea was for whitelisting of routes for legal rep which exludes routes not setup in the router such as "dashboard" legal rep users shouldn't be able to access this

paddy-hmcts (approved):

PR #1057, opened 2026-04-10
HDPI-5635-other-considerations-screen

AlexBuzea:
### Jira link

See [HDPI-5635](https://tools.hmcts.net/jira/browse/HDPI-5635)

### Change description

-adding `Other considerations` page and a placeholder page for the one after `Upload a file`

image
image

### Testing done

-tested UI manually to make sure works as expected

  * [ ] Yes
  * [x] No

### Checklist


src/main/assets/locales/cy/respondToClaim/otherConsiderations.json
```diff
@@ -0,0 +1,9 @@
+{
+  "pageTitle": "cyOther considerations",
+  "heading": "cyIs there anything else you want to tell the court as part of your response?",
+  "otherConsiderationsDetailsLabel": "Give details about what else you want to tell the court",
```
scottstewart-sl: Prefix with "cy" ?

src/main/steps/respond-to-claim/other-considerations/index.ts
```diff
@@ -0,0 +1,97 @@
...
+  },
+  beforeRedirect: async req => {
+    const otherConsiderations: YesNoValue | undefined = req.body?.otherConsiderations;
+
+    if (!otherConsiderations) {
+      return;
+    }
+
+    const otherConsiderationsDetails: string | undefined =
+      otherConsiderations === 'Yes' ? req.body?.['otherConsiderations.otherConsiderationsDetails'] : undefined;
```
scottstewart-sl: Could you just default to an empty string here and then not have to do it on L31 below? Having said that, shouldn't we be sending null in the JSON rather than an empty string for fields that are not relevant?

src/main/steps/respond-to-claim/other-considerations/index.ts
```diff
@@ -0,0 +1,97 @@
...
+  customTemplate: `${__dirname}/otherConsiderations.njk`,
+  translationKeys: {
+    pageTitle: 'pageTitle',
+    heading: 'heading',
+  },
+  beforeRedirect: async req => {
+    const otherConsiderations: YesNoValue | undefined = req.body?.otherConsiderations;
+
+    if (!otherConsiderations) {
+      return;
```
scottstewart-sl, line 27: If this condition matches, does that mean that something has gone wrong? Perhaps it should at least log a warning.

src/main/steps/respond-to-claim/other-considerations/index.ts
AlexBuzea: Last I looked at this in a previous PR the backend doesn't accept null values and this empty string is needed to clear any previous values if the user changes his answer from 'yes' to 'no'. I added the check on line 31 because it feels more defensive but can change it if you feel strongly about it

src/main/steps/respond-to-claim/other-considerations/index.ts
AlexBuzea: I've added a warning with it now thanks

src/main/assets/locales/cy/respondToClaim/otherConsiderations.json
AlexBuzea: Thanks, missed that one

src/main/steps/respond-to-claim/other-considerations/index.ts
scottstewart-sl: The backend does accept null values but won't clear the previous draft values as you say. (There is a PR coming up that will change the way the drafts are saved which should resolve this)

scottstewart-sl (approved):

libanAbdirahman1 (approved):

PR #1065, opened 2026-04-13
HDPI-4268: Gen apps notice and sharing with other parties screens

scottstewart-sl:
### Jira link

See [HDPI-4268](https://tools.hmcts.net/jira/browse/HDPI-4268)

### Change description

Add two screens to the citizen Gen App journey for Without Notice and Sharing. I've also added a new class `VisibleFormDataView` which wraps the request form data from the journey and exposes the answers from the journey that are applicable based on the step and field show conditions. This is used on the placeholder CYA page which is still being developed and will be fully implemented in a later ticket.

Corresponding pcs-api PR: https://github.com/hmcts/pcs-api/pull/1654

have-the-other-parties-agreed-to-this-application

are-there-any-reasons-that-this-application-should-not-be-shared

### Testing done

Manual testing locally with corresponding changes in pcs-api. Unit tests updated.

  * [ ] Yes
  * [x] No

### Checklist


src/main/steps/make-an-application/check-your-answers/visibleFormDataView.ts
```diff
@@ -0,0 +1,124 @@
...
+    }
+
+    return {
+      stepName,
+      fieldValue: getFormData(this.req, stepName).typeOfApplication as GenAppType,
+    };
+  }
+
+  getHearingInNext14DaysField(): FieldDetails<'yes' | 'no'> | undefined {
+    const stepName = 'is-the-court-hearing-in-the-next-14-days';
```
paddy-hmcts: as this is a hardcoded value, you could move it to being a param in the function with a default value, promoting reuse

src/main/steps/make-an-application/check-your-answers/visibleFormDataView.ts
```diff
@@ -0,0 +1,124 @@
...
+
+    if (reasonsAppShouldNotBeShared === 'yes') {
+      return {
+        stepName,
+        fieldValue: getFormData(this.req, stepName)['reasonsAppShouldNotBeShared.reasonForNotSharing'] as string,
+      };
+    } else {
+      return undefined;
+    }
+  }
```
paddy-hmcts, line 76: I know the functions in here are specific, but they have a lot of 'duplicated' behaviour/return types etc. You could probably abstract this further and make use of generics (see https://www.typescriptlang.org/docs/handbook/2/generics.html)

paddy-hmcts (approved):

src/main/steps/make-an-application/check-your-answers/visibleFormDataView.ts
scottstewart-sl: Now updated to have a shared function with generics

src/main/steps/make-an-application/check-your-answers/visibleFormDataView.ts
scottstewart-sl: As mentioned in the chat, I don't think there would be any value in making this a function parameter because the step name is linked to the field that is being fetched, so any other step name wouldn't find anything, (unless there was another field with the same name on a different page, but these field names are quite specific)

src/main/steps/make-an-application/check-your-answers/visibleFormDataView.ts
```diff
@@ -0,0 +1,77 @@
...
+export type FieldDetails<T> = {
+  stepName: string;
+  fieldValue: T;
+};
+
+export default class VisibleFormDataView {
+  constructor(readonly req: Request) {}
+
+  getApplicationTypeField(): FieldDetails<GenAppType> | undefined {
+    return this.getField('choose-an-application', 'typeOfApplication');
```
paddy-hmcts, line 18: You should probably pass in the type to the `getField` function too, e.g.:

```
getApplicationTypeField(): FieldDetails<GenAppType> | undefined {
    return this.getField<FieldDetails<GenAppType> | undefined>('choose-an-application', 'typeOfApplication');
```

src/main/steps/make-an-application/check-your-answers/visibleFormDataView.ts
```diff
@@ -0,0 +1,77 @@
...
+import { flowConfig } from '../flow.config';
+
+import { getFormData } from '@modules/steps';
+
+export type FieldDetails<T> = {
+  stepName: string;
+  fieldValue: T;
+};
+
+export default class VisibleFormDataView {
```
linusnorton, line 14: I'm a little confused on the purpose of this class. It feels like it's just a way to add type information to `req`? Why not just do that and then you can just access these fields with `req['my-step']['my-field']`

linusnorton: I see we're mixing in some logic for removing fields in `getReasonForNotSharingField` - is this the right place for that? Shouldn't that happen the way Arun is approaching it?

src/main/steps/showConditionService.ts
```diff
@@ -0,0 +1,12 @@
+import type { Request } from 'express';
+
+import { JourneyFlowConfig } from '../interfaces/stepFlow.interface';
+
+export function shouldShowStep(req: Request, stepName: string, flowConfig: JourneyFlowConfig): boolean {
```
linusnorton: Is this worth it's own file or should it live in the index.ts with the other step management functions?

It almost feels like this and the index should be pulled out to a module somewhere.

src/main/steps/make-an-application/check-your-answers/visibleFormDataView.ts
scottstewart-sl: Yes, I was going to do that but it compiles without it, (and breaks if I deliberately put the wrong type), so I suspect it is inferring the type based on the return value.

src/main/steps/make-an-application/check-your-answers/visibleFormDataView.ts
scottstewart-sl: This class is a view onto the request form data that applies the step show condition, and optionally the field show condition if there is one. I wanted a single place to encapsulate those rules so that the CYA page doesn't need to concern itself with them and can just display and submit a field value if this view provided it.

The way that Arun is approaching it is for a journey where a draft is being saved page by page and the page post handler is responsible for determining which fields on that page should be sent to the pcs-api. This journey doesn't have draft functionality yet and works with the request form data from the session at the end of the journey.

Also a subsequent change to an earlier answer can mean that a later page is no longer relevant/visible, but the post handler wouldn't be called again for that page and hence the now-hidden data wouldn't be cleared.

src/main/steps/showConditionService.ts
scottstewart-sl: Yes, I'll move it into the index.ts with the other functions.

scottstewart-sl: Now updated

linusnorton (approved):

PR #1082, opened 2026-04-14
HDPI-5701: RTC Sections (plumbing)

andrewdami11:
### Jira link

See [HDPI-5701](https://tools.hmcts.net/jira/browse/HDPI-5701)

### Change description

This change adds section plumbing for the respond-to-claim journey, so future [task-list](https://tools.hmcts.net/jira/browse/HDPI-5350) and [section CYA](https://tools.hmcts.net/jira/browse/HDPI-5407) work can use this model.

The branch introduces a dedicated section config for respond-to-claim, including ordered section membership, translation title keys, and conditional applicability for the payments section using CCD-backed case data
(`hasAnyRentArrearsGround`).

It also adds shared section utilities for resolving section membership, section entry steps, section applicability, and
detecting section-boundary transitions from current/next step values.

RTC step ordering is centralised in a dedicated `stepOrder` module, and `stepRegistry` is
type-constrained against that step list (excluding terminal `end-now`) to reduce duplication drift.

Related pcs-api PR: https://github.com/hmcts/pcs-api/pull/1759

### Testing done

- `yarn lint`
- `yarn test:unit`
- `yarn test:unit src/test/unit/steps/utils/sections.test.ts src/test/unit/steps/respond-to-claim/sections.config.test.ts`
- `yarn test:unit src/test/unit/routes/registerSteps.test.ts src/test/unit/steps/respond-to-claim/sections.config.test.ts src/test/unit/app/utils/stepFlow.test.ts`
- `yarn build`

Added/updated unit coverage:

- `src/test/unit/steps/utils/sections.test.ts`
- `src/test/unit/steps/respond-to-claim/sections.config.test.ts`
- `src/test/unit/routes/registerSteps.test.ts`

-   [ ] Yes
-   [x] No

### Checklist

-   [x] commit messages are meaningful and follow good commit message guidelines
-   [ ] README and other documentation has been updated / added (if needed)
-   [x] tests have been updated / new tests has been added (if needed)
-   [ ] Does this PR introduce a breaking change


src/main/steps/respond-to-claim/sections.config.ts
```diff
@@ -0,0 +1,91 @@
+import type { SectionConfig } from '../../interfaces/stepFlow.interface';
+import { hasAnyRentArrearsGround } from '../utils';
+
+export const respondToClaimSections = {
+  startNowAndDetails: {
+    titleKey: 'taskList.startNowAndDetails',
+    order: 1,
```
scottstewart-sl, line 7: Rather than having an order field, could `respondToClaimSections` just be an array?

src/main/steps/respond-to-claim/sections.config.ts
```diff
@@ -0,0 +1,91 @@
...
+    steps: ['start-now', 'free-legal-advice'],
+  },
+  personalDetails: {
+    titleKey: 'taskList.personalDetails',
+    order: 2,
+    steps: [
+      'defendant-name-confirmation',
```
scottstewart-sl, line 14: The step names for a journey are now in up to 3 places - is the `stepOrder` field in `flow.config.ts` still needed if the order is defined here? I think we should be aiming to reduce the duplication where possible.

Also what is there to stop a routing condition for a step in one section routing to a step in a different section?

src/main/steps/respond-to-claim/sections.config.ts
andrewdami11: I set it up this way because `order` is explicit and future-proof, in the case that sections are conditionally omitted/reordered

src/main/steps/respond-to-claim/sections.config.ts
andrewdami11: Reduced duplication [here](https://github.com/hmcts/pcs-frontend/pull/1082/commits/856dc2871726d204685bfac6cb39d4311e26d821)

Wrt _mid-section crossovers_, I think it’d be out of scope. This PR is plumbing only, not behaviour enforcement. 
I don’t think we have there's an explicit product decision/direction yet to globally block mid-section crossovers, so I wouldn't introduce that guardrail here yet.

libanAbdirahman1 (approved):

src/main/steps/respond-to-claim/sections.config.ts
scottstewart-sl: I don't think the duplication has been reduced, just that the steps in `stepOrder` have been moved out to another file. What is `stepOrder` actually used for in a sectionalised journey?

src/main/steps/respond-to-claim/sections.config.ts
scottstewart-sl: Is reordering likely to ever be a requirement? And if you are conditionally hiding a whole section, then isn't that along the same lines as hiding a single step, i.e. you just move to the following item in the array when iterating.

Having an explicit section order property here makes it different to how the step order is managed. Also there isn't anything to stop someone accidentally put the same order number twice or deleting an section and not updating the order on the following sections.

Maybe a compromise is to have the map of sections keyed on section ID, like you have now, and a separate array of the section IDs. That would be consistent with how the steps are ordered.

Anyway, I couldn't see where the order property is actually used in the code, (perhaps it will be on the case dashboard?). So maybe it can be completely removed for now and only added back in when it is known how the code will be using it.

andrewdami11: Fairs, removed `order` [here](https://github.com/hmcts/pcs-frontend/pull/1082/commits/5f816951fe3f0aa39197bac4f29c0d9a21a29f61)

src/main/steps/respond-to-claim/sections.config.ts
andrewdami11: Refactored flow.config's `stepOrder` [here](https://github.com/hmcts/pcs-frontend/pull/1082/commits/37fc3b4bff425c8347d307fc07ad2b531b1ce7fb) now

scottstewart-sl (approved):

PR #1089, opened 2026-04-15
Add local vault for secrets

linusnorton:
Add local vault loading for secrets when working with AAT locally.

Running: `yarn start:dev` will load the vault secrets and connect to AAT.
Running: `yarn start:dev:pcs-local` will load the local environment configuration and not get secrets from the vault.

Note: this change will require developers wanting to connect to AAT to have logged in with `az login`


package.json
```diff
@@ -7,9 +7,10 @@
   },
   "scripts": {
     "docker:login:hmctsprod": "az acr login --name hmctsprod",
-    "start:dev:pcs-local": "NODE_ENV=development PCS_API_URL=http://localhost:3206 CCD_URL=http://localhost:4452 S2S_URL=http://localhost:8489 OIDC_ISSUER=http://localhost:5062/o OIDC_CLIENT_ID=pcs-frontend OIDC_REDIRECT_URI=http://localhost:3209/oauth2/callback XUI_BASE_URI=http://localhost:3000 yarn start:dev",
+    "start:dev:pcs-local": "USE_VAULT=false NODE_CONFIG_ENV=pcs-local yarn start:dev",
     "start:dev:ssl": "./bin/generate-ssl-options.sh && NODE_ENV=development USE_SSL=true OIDC_REDIRECT_URI=https://localhost:3209/oauth2/callback yarn start:dev",
-    "start:dev": "node --trace-deprecation --env-file-if-exists=.env -r ts-node/register -r tsconfig-paths/register src/main/bootstrap.ts",
+    "redis:up": "docker start pcs-redis 2>/dev/null || docker run -d --name pcs-redis -p 6379:6379 redis || echo '[redis:up] Could not start Redis via Docker. App will fail to connect unless Redis is already running on localhost:6379.'",
+    "start:dev": "yarn redis:up && node --trace-deprecation --env-file-if-exists=.env -r ts-node/register -r tsconfig-paths/register src/main/bootstrap.ts",
```
scottstewart-sl: There's also a wiremock server to run for the FeeRegister stubs, (still in on a [PR branch](https://github.com/hmcts/pcs-frontend/blob/HDPI-4267_apply_something_else/docker-compose-dependencies-only.yml) at the moment). So maybe it is better to start Redis in the docker compose along with Wiremock and perhaps the s2s server?

config/pcs-local.json
```diff
@@ -0,0 +1,17 @@
+{
+  "api": {
+    "url": "http://localhost:3206"
+  },
+  "ccd": {
+    "url": "http://localhost:4452"
+  },
+  "s2s": {
+    "url": "http://localhost:8489"
```
scottstewart-sl, line 9: Does the pcs-api bootWithCCD actually start an s2s service? I thought it just intercepted calls and generated a fake token for local dev? In my experience, I've had to start a separate s2s container for local frontend dev. (Related to my other comment below about docker compose)

package.json
linusnorton: Can do, but I think the service itself is also in that docker-compose file... does anyone actually use it that way? I would be include to take the frontend out of docker-compose.yml and just use it for the dev deps

config/pcs-local.json
linusnorton: It does, it's part of cft lib and not idam simulator

scottstewart-sl: I had a different port configured for some reason, (4502).

scottstewart-sl (approved):

adusumillipraveen (approved):

PR #1130, opened 2026-04-17
Improved file locks for reducing pipeline execution time

Gautham059:
### Jira link

https://tools.hmcts.net/jira/browse/HDPI-6103

### Change description

1. Replaced single pft-locks directory with pft-locks/passed and pft-locks/failed subdirectories to separately track pass/fail outcomes per page
2. Failed lock is written immediately when any of the 4 functional test types (content validation, EMV, navigation, visibility) fails for a page; passed lock is written only when all applicable types complete without failure
3. Pages with a failed lock are always re-run on a new test case or retry, but skipped if already visited in the same test run - preventing both duplicate execution and permanent suppression of known failures
4. Pages with a passed lock are skipped entirely across runs until a failure is detected, at which point the passed lock is deleted and a failed lock is created
5. In-memory guard (pagesTestedInCurrentRun) is now checked first in all cases - including excluded and failed-lock pages - so no page runs functional tests twice within the same test run regardless of its lock state
6. Added lock-exclusions.config.ts support: excluded pages always re-run on a fresh visit but still write pass/fail locks like any other page
7. Added PageContentValidation.hasErrorForPage() and a dedicated pagesWithErrors set so content validation failures are detectable immediately within the trigger file, consistent with how EMV, navigation and visibility failures are handled

### Testing done

Tested all negative scenarios in this build - https://build.hmcts.net/job/HMCTS_j_to_z/job/pcs-frontend/job/PR-1130/24/Frontend_20Functional_20Test_20Report/ (Introduced all possible issues, our framework could catch those successfully without repeating the scenarios)
image

passed build - 
image


srinijg (approved):
LGTM

PR #1133, opened 2026-04-18
HDPI-6053: RTC ShowCondition routing

andrewdami11:
### Jira link

See [HDPI-6053](https://tools.hmcts.net/jira/browse/HDPI-6053)

### Change description

This branch completes the respond-to-claim move to show-condition navigation so step visibility is driven from one model.

For respond-to-claim, the flow config now relies on `stepOrder` and per-step `showCondition` checks, with no step-level `routes`, `defaultNext`, or `previousStep` entries in the journey config. This keeps routing behaviour consistent and reduces URL deep-link relevance issues.

Citizen and legal representative variants remain aligned by sharing the same show-condition-based step definitions, with legal-rep-specific step ordering retained where needed.

### Testing done

- Updated unit tests:
  - `src/test/unit/steps/respond-to-claim/navigation-from-case-data.test.ts`
  - `src/test/unit/steps/utils/hasRentArrearsGrounds.test.ts`
- Executed locally:
  - `yarn lint` (pass)
  - `yarn test:unit` (pass)
  - `yarn build` (pass)

-   [ ] Yes
-   [x] No

### Checklist

-   [x] commit messages are meaningful and follow good commit message guidelines
-   [ ] README and other documentation has been updated / added (if needed)
-   [x] tests have been updated / new tests has been added (if needed)
-   [ ] Does this PR introduce a breaking change


arun-hmcts (commented):
@andrewdami11  I think the intention here was to follow the same show-condition setup introduced earlier on master branch for gen-app (src/main/steps/make-an-application/flow.config.ts) — useShowConditions: true with stepOrder and per-step sync showCondition, without routes, defaultNext, or previousStep.
In this PR, it feels like showCondition has been added alongside the existing routing rather than replacing it, so each step ends up carrying both. Just wondering if we’d want to simplify that and rely on one approach.
Also, I’m not sure we need the widened ShowCondition signature (async (req, formData, currentStepData)) or the useSessionFormData flag. Gen-app reads from session via getFormData, but for respond-to-claim, CCD case data is the source of truth. Since caseReferenceParamMiddleware refreshes res.locals.validatedCase on every request, the conditions could just read from there.

scottstewart-sl:
I agree with keeping the method signature clean for the show conditions. Hopefully the `req` should have everything the show condition needs - either the `validatedCase` or the `formData` depending on requirements. Also is there anything that actually needs it to be async at the moment?

src/main/modules/steps/flow.ts
```diff
@@ -257,6 +257,19 @@ export function stepDependencyCheckMiddleware(flowConfigOrResolver: JourneyFlowC
...
+      const canShowStep = flowConfig.steps[stepName].showCondition(req);
+      if (!canShowStep) {
+        logger.debug(`Step ${stepName} is hidden by showCondition`);
+        const caseReference = req.res?.locals.validatedCase?.id;
+        const journeyStartStep = await getFirstVisibleStep(req, flowConfig);
+        if (journeyStartStep) {
+          return res.redirect(303, getStepUrl(journeyStartStep, flowConfig, caseReference));
+        }
+      }
+    }
```
scottstewart-sl: Although we will need this sort of check, I don't think it was part of the ticket and as far as I know it hasn't been agreed what the behaviour should be if the user does try to access a page that isn't visible. It would be better to keep on the side for now, and not add complexity to this PR.

src/main/steps/respond-to-claim/flowConditions.ts
```diff
@@ -0,0 +1,70 @@
...
+export function getContactByTelephoneAnswer(req: Request): 'yes' | 'no' | undefined {
+  const fromCcd = req.res?.locals?.validatedCase?.isDefendantContactByPhone;
+  if (fromCcd === true) {
+    return 'yes';
+  }
+  if (fromCcd === false) {
+    return 'no';
+  }
+
+  return undefined;
```
scottstewart-sl: Minor - I know this has just been moved out, but this could be simplified for the show condition. It is getting a `boolean` from the `validatedCase` mapping it to 'yes' or 'no' and then checking if the value is 'yes' in the show condition, which converts it back to a boolean. Probably you don't need it at all, and the show condition can simply become something like:

```
(req: Request) => req.res?.locals?.validatedCase?.isDefendantContactByPhone
```

(or just wrap that one line in a function if you want to keep it that way).

src/main/steps/respond-to-claim/flowConditions.ts
```diff
@@ -0,0 +1,70 @@
...
+      ?.repayArrearsInstalments;
+  return ccdAnswer === 'YES';
+}
+
+export function shouldShowInstallmentPaymentsStep(req: Request): boolean {
+  return hasRejectedRepaymentAgreement(req) && hasAnyRentArrearsGround(req);
+}
+
+export function shouldShowHowMuchAffordToPayStep(req: Request): boolean {
+  return hasConfirmedInstallmentOffer(req);
```
scottstewart-sl: Minor - you could just inline this, then it would be clearer in the flow what the show condition was

src/main/steps/utils/normalizeYesNoNotSureValue.ts
```diff
@@ -0,0 +1,22 @@
...
+  }
+
+  const normalized = value.trim().toUpperCase();
+  if (normalized === 'YES') {
+    return 'yes';
+  }
+  if (normalized === 'NO') {
+    return 'no';
+  }
+  if (normalized === 'NOT_SURE' || normalized === 'IM_NOT_SURE' || normalized === 'IMNOTSURE') {
```
scottstewart-sl: Where are these variations coming from? We should use `NOT_SURE` which is the one in the CCD model.

Also this seems to overlap quite heavily with `defendantResponsesConfirmNoticeGiven` in `ccdCaseData.model.ts`. Is there some duplication that could be removed?

src/main/steps/respond-to-claim/flowConditions.ts
```diff
@@ -0,0 +1,70 @@
...
+  if (fromCcd === false) {
+    return 'no';
+  }
+
+  return undefined;
+}
+
+export function getConfirmNoticeGivenAnswer(req: Request): 'yes' | 'no' | 'imNotSure' | undefined {
+  const ccdAnswer = req.res?.locals?.validatedCase?.defendantResponsesConfirmNoticeGiven;
+  if (ccdAnswer === 'yes' || ccdAnswer === 'no' || ccdAnswer === 'imNotSure') {
```
scottstewart-sl: Probably not for this PR, but maybe `defendantResponsesConfirmNoticeGiven` should be returning CCD values, (YES/NO/NOT_SURE). It seems to be converting the value to what it would have been when it was coming from the session form data, so maybe that's leftover from before.

src/main/steps/respond-to-claim/flowConditions.ts
```diff
@@ -0,0 +1,70 @@
...
+    return ccdAnswer;
+  }
+
+  const ccdNestedAnswer = normalizeYesNoNotSureValue(
+    req.res?.locals?.validatedCase?.data?.possessionClaimResponse?.defendantResponses?.confirmNoticeGiven ??
+      req.res?.locals?.validatedCase?.data?.possessionClaimResponse?.defendantResponses?.possessionNoticeReceived
```
scottstewart-sl: Was this existing code? If `confirmNoticeGiven` was defined wouldn't it have been handled further up this function? I'm also not sure what is "nested" about the `possessionNoticeReceived` field - for my benefit could you clarify the difference between `confirmNoticeGiven` and `possessionNoticeReceived` ?

src/main/steps/utils/normalizeYesNoNotSureValue.ts
andrewdami11: Oh my bad, those are stale changes, from my local _devving_. I'll remove them

src/main/steps/respond-to-claim/legalrep.flow.config.ts
arun-hmcts, line 1: Legal representative is a separate journey with its own steps and ordering, and it’s currently not moving to the task list model like the citizen journey (still undecided). Given that, is there a specific reason we’re proceeding with this refactoring?

src/main/steps/respond-to-claim/legalrep.flow.config.ts
```diff
@@ -1,416 +1,9 @@
...
-    'what-other-regular-expenses-do-you-have': {
-      previousStep: 'priority-debt-details',
-      defaultNext: 'end-now',
-    },
-    'installment-payments': {
-      previousStep: 'repayments-agreed',
-      defaultNext: 'your-household-and-circumstances',
-    },
-  },
+  stepOrder: citizenFlowConfig.stepOrder.filter(step => step !== 'free-legal-advice'),
```
arun-hmcts: Filtering citizen's stepOrder can't add new steps or reorder, so this won't hold once legal rep diverges.

src/main/steps/respond-to-claim/legalrep.flow.config.ts
andrewdami11: I thought we agreed to migrate legal-rep as well to use showCondition, in our 2pm call on the 15th of April :/

andrewdami11: Clarified [here](https://hmcts-reform.slack.com/archives/C0ATVP9JJ4Q/p1776766548523999?thread_ts=1776766054.108709&cid=C0ATVP9JJ4Q). Going to _Resolve_ this comment

src/main/modules/steps/flow.ts
andrewdami11: Addressed [here](https://github.com/hmcts/pcs-frontend/pull/1133/commits/bd32523f01eefcef177011a8527124b88ba8b89f)

src/main/steps/respond-to-claim/flowConditions.ts
andrewdami11: Addressed [here](https://github.com/hmcts/pcs-frontend/pull/1133/commits/bd32523f01eefcef177011a8527124b88ba8b89f)

src/main/steps/respond-to-claim/flowConditions.ts
andrewdami11: Addressed [here](https://github.com/hmcts/pcs-frontend/pull/1133/commits/13d6d5cc1a8cfc4a8e7e8eed8d040c8a47673d29)

src/main/steps/utils/normalizeYesNoNotSureValue.ts
andrewdami11: Addressed [here](https://github.com/hmcts/pcs-frontend/pull/1133/commits/bd32523f01eefcef177011a8527124b88ba8b89f)

src/main/steps/respond-to-claim/legalrep.flow.config.ts
andrewdami11: Good point, addressed [here](https://github.com/hmcts/pcs-frontend/pull/1133/commits/76a26331731df32c5b08093a1498dc24bc70ce49)

src/main/steps/respond-to-claim/flowConditions.ts
andrewdami11: Addressed [here](https://github.com/hmcts/pcs-frontend/pull/1133/commits/0e9250b5fab7a4d982a5af6d0248b6792a53bc09)

src/main/steps/respond-to-claim/flowConditions.ts
andrewdami11: Simplified things [here](https://github.com/hmcts/pcs-frontend/pull/1133/commits/0e9250b5fab7a4d982a5af6d0248b6792a53bc09)

src/main/steps/utils/normalizeYesNoNotSureValue.ts
```diff
@@ -0,0 +1,22 @@
+export function normalizeYesNoNotSureValue(value: unknown): 'yes' | 'no' | 'imNotSure' | undefined {
```
scottstewart-sl: Is this used anywhere now?

andrewdami11: Addressed [here](https://github.com/hmcts/pcs-frontend/pull/1133/commits/cd8854e9c1c11cacebde5c78a6d92aca9f6b6f3f)

scottstewart-sl (approved):

arun-hmcts (approved):

PR #1144, opened 2026-04-21
suppress cve

linusnorton:
As far as I can tell we don't actually have the 8 series of protobuf in our repo, everything should be 7.x so I'm not even sure why this is flagged.


scottstewart-sl:
Can you try upgrading to 7.5.5 anyway to see if that also resolves the warning?

PR #1274, opened 2026-05-05
Hdpi 6221 select defendant

m014389b:
### Jira link

See [HDPI-6221](https://tools.hmcts.net/jira/browse/HDPI-6221)

### Change description
Introduce select defendant screen which will display when multiple defendants are linked to the legal representative otherwise will skip the screen and go to the defendant name confirmation screen. 

Introduced clientContextHeaders to hold the "selectedPartyId" and will pass this to the start, submit and page mid events to retrieve the correct draft. 

Clear context header on start now screen to prevent case data apis being invoked with other cases selected party IDs

### Testing done

Unit and manual testing

  * [ ] Yes
  * [x] No

### Checklist


adusumillipraveen (approved):

src/main/services/ccdCase.interface.ts
```diff
@@ -115,6 +124,11 @@ export interface CcdClaimGroundSummaryItem {
   id: string;
 }
 
+export interface CcdDefendantItem {
+  value: CcdDefendantParty;
+  id: string;
+}
```
scottstewart-sl, line 122: Instead of making a new type for this, can you use `CcdCollectionItem` with a generic type?

src/main/services/ccdCase.interface.ts
```diff
@@ -21,6 +21,15 @@ export type EqualityAndDiversityQuestionsChoice = 'CONTINUE' | 'SKIP' | null;
...
 
+export type RadioItems = {
+  component: {
+    items: {
+      text: string;
+      value: string;
+    }[];
+  };
+};
+
```
scottstewart-sl: I'm not sure if this is the best place for this type - it's not a CCD type / collection is it?

src/main/steps/utils/isSingleLinkedDefendant.ts
```diff
@@ -0,0 +1,7 @@
+import type { Request } from 'express';
+
+export function isSingleLinkedDefendant(req: Request): boolean {
+  const caseData = req.res?.locals?.validatedCase?.data;
+
+  return caseData?.allDefendants?.length === 1;
```
scottstewart-sl: Hmmm, I think the `allDefendants` collection should always have all the defendants in it, not just those represented by the LR. I think it might be better to represent that list as a dedicated field in the event DTO

src/main/middleware/clientContextSessionClearer.ts
```diff
@@ -0,0 +1,18 @@
...
+  const segments = path.split('/').filter(Boolean);
+
+  if (segments[0] === 'case' && segments.length > 3) {
+    return `${segments.slice(3).join('/')}`;
+  }
+  return path;
+};
+
+export const clientContextSessionClearerMiddleware = (req: Request, res: Response, next: NextFunction): void => {
+  const setpPath = getStepPath(req.path);
```
scottstewart-sl: Minor - typo in `stepPath`

src/main/middleware/clientContextSessionClearer.ts
```diff
@@ -0,0 +1,18 @@
...
+  if (setpPath.includes('start-now')) {
```
scottstewart-sl: This feels slightly brittle - it's not guaranteed that the first page in each journey will have the step name `start-now`. How about clearing the session property directly in each of the actual first pages in the journey.

Also does this need to be namespaced by the journey name? I wonder if a user could have two different journeys on the go at the same time

src/main/steps/respond-to-claim/select-defendant/index.ts
```diff
@@ -0,0 +1,67 @@
...
+    const allLinkedDefendants: CcdCollectionItem<CcdDefendantParty>[] | undefined =
+      req.res?.locals?.validatedCase?.allLinkedDefendants;
+
+    const radio = formContent.fields.find(f => f.componentType === 'radios') as RadioItems | undefined;
+
+    addRadioButtonForAllLinkedDefendants(allLinkedDefendants, radio);
+
+    return {
+      ...formContent,
+    };
```
scottstewart-sl: I'm not an expert at TS/JS, but out of interest is this different to just putting `return formContent` ?

src/main/steps/respond-to-claim/select-defendant/index.ts
```diff
@@ -0,0 +1,67 @@
...
+    pageTitle: 'pageTitle',
+    caption: 'caption',
+    contactUs: 'contactUs',
+    detailsHeading: 'detailsHeading',
+  },
+  extendGetContent: async (req, formContent) => {
+    const allLinkedDefendants: CcdCollectionItem<CcdDefendantParty>[] | undefined =
+      req.res?.locals?.validatedCase?.allLinkedDefendants;
+
+    const radio = formContent.fields.find(f => f.componentType === 'radios') as RadioItems | undefined;
```
scottstewart-sl: Maybe this should find by field name rather than component type. At the moment it is fine because there is only one radio component on the form, but field name should always be unique even if another radio option was added.

src/main/middleware/clientContextSessionClearer.ts
m014389b: done

src/main/steps/respond-to-claim/select-defendant/index.ts
m014389b: Good point

src/main/steps/respond-to-claim/select-defendant/index.ts
m014389b: Correct, I was spreading it but was pointless

src/main/middleware/clientContextSessionClearer.ts
m014389b: moved to start-now postController.

scottstewart-sl (approved):

PR #1276, opened 2026-05-05
view documents screen

adusumillipraveen:
### Jira link

https://tools.hmcts.net/jira/browse/HDPI-5549 

### Change description

- adds view-documents screen
- uses generic getcase from ccd to get the document list

### Testing done

Tested with https://github.com/hmcts/pcs-api/pull/1779

### Checklist


src/main/services/ccdCaseService.ts
```diff
@@ -174,7 +174,11 @@ async function submitEvent(
 }
 
 export const ccdCaseService = {
-  async getCaseById(accessToken: string, caseId: string, eventId: string = 'respondPossessionClaim'): Promise<CcdCase> {
+  async getCaseByIdForEvent(
```
scottstewart-sl, line 177: I'm making similar changes in my branch - I called this one `startEvent` because that's what it is actually calling. It depends if you want to hide the fact that we are using startEvent to get specific case data sometimes.

src/main/utils/documentUtils.ts
```diff
@@ -0,0 +1,114 @@
...
+export function formatSubmittedOn(
+  dateValue?: string,
+  locale = 'en-GB',
+  submittedOnPrefix = 'Submitted on'
+): string | null {
+  if (!dateValue) {
+    return null;
+  }
+
+  const date = new Date(normaliseDateValue(dateValue));
```
scottstewart-sl: Could you use Luxon for this date formatting? There's even a Nunjucks date filter that we've added that you may be able to use directly in the template, like the example below. The default format is probably the one you want, and it handles the conversion into the UK timezone

```
<p class="govuk-body-s">{{ t('viewAllApplications:submittedOn') }} {{ genApp.submittedOn | date}}</p>
```

src/main/services/ccdCaseService.ts
adusumillipraveen: yeah, i wanted to keep it `getCaseXXX` on purpose  to make it agnostic.

src/main/utils/documentUtils.ts
adusumillipraveen: didn't knew that, thanks.

scottstewart-sl (approved):

src/main/services/ccdCaseService.ts
```diff
@@ -198,6 +202,29 @@ export const ccdCaseService = {
     }
   },
 
+  async getCaseById(accessToken: string, caseId: string): Promise<CcdCase> {
+    const caseUrl = `${getBaseUrl()}/cases/${caseId}`;
+
+    try {
+      logger.debug(`[ccdCaseService] Fetching case by id forread view: ${caseId}`);
+      const response = await http.get<CcdCase>(caseUrl, getCaseHeaders(accessToken));
+      logger.debug(`[ccdCaseService]read case response for ${caseId}: ${JSON.stringify(response.data, null, 2)}`);
```
scottstewart-sl: Btw, I'm not familiar with this logger, but it seems odd that we have to prefix the log messages with what is effectively the logger name, when it has already been provided in the call to `getLogger`. Should this be something that is added to the logger format?

fudge88 (approved):

SameenaHMCTS (approved):
lgtm

PR #1284, opened 2026-05-07
HDPI-5181-file-upload-screens

sdmiddletonSolirius:
### Jira link

See [HDPI-5181](https://tools.hmcts.net/jira/browse/HDPI-5181)

### Change description

Backend PR: https://github.com/hmcts/pcs-api/pull/1792

### **2 Screens added**
- /counter-claim-do-you-want-to-upload-files
- /counter-claim-upload-files

### **/counter-claim-do-you-want-to-upload-files**
- Implemented yes/no radio selection which maps to new `counterClaimWantToUploadFiles": "YES"` field in draft_case_data

Screenshot 2026-05-14 at 17 03 34

Screenshot 2026-05-14 at 17 07 25

### **/counter-claim-upload-files**
- Extended the work carried out for upload-documents page
- Data persists to docstore and docstore urls successfully persisted to draft_case_data
Screenshot 2026-05-14 at 17 07 32

### **Bug fix: file upload component breaking when "Save and continue" was clicked without uploading a file**

- `counter-claim-upload-files` screen is the first upload step in the journey that requires at least one file before the user can proceed - a user could click "Save and continue" without uploading anything and the journey would proceed. 
- Validation rule added to make uploading at least one file mandatory before the user can continue (with a bypass for "Save for later"). 
- When a user triggered the error by clicking "Save and continue" without a file, the MOJ drag-and-drop file upload component would visually break, reverting to a plain browser file input and displaying a long list of empty "Remove" buttons (see screenshot)
remove-buttons-break

- when a page is re-displayed after a validation failure, the framework was not re-attaching the AJAX upload and delete endpoint URLs to the file upload component. Without those URLs, the MOJ JavaScript enhancement could not initialise, causing the visual degradation. The existing upload-document screen (which has no mandatory upload rule) had never triggered this code path and so the gap had gone unnoticed until now.
- new module `src/main/modules/steps/formBuilder/fileUploadUtils.ts` was introduced containing three utility functions. `wireFileUploadUrls` re-attaches the correct upload and delete endpoint URLs to the file component - this was previously only happening on a normal page load (GET), and `fileUploadUtils` now ensures it also runs when a page is re-rendered after a validation error (POST). `wireFileUploadOnPostError` builds on this by additionally restoring any files the user had already uploaded to the visible list, using `parseUploadedDocumentsFromBody` to reconstruct them from the hidden inputs submitted with the form. These utilities are called from `postHandler.ts` during the error re-render path, which now receives the step's `documentStorage` configuration - the key change that allows the framework to know a file upload step is involved and apply the fix.

Screenshot 2026-05-14 at 17 03 47

**Error messages**
Screenshot 2026-05-14 at 17 03 59
Screenshot 2026-05-14 at 17 04 26
Screenshot 2026-05-14 at 17 06 15

  * [ ] Yes
  * [X] No

### Checklist


src/main/assets/locales/cy/respondToClaim/counterClaimDoYouWantToUploadFiles.json
```diff
@@ -0,0 +1,13 @@
+{
+  "pageTitle": "cyDo you want to upload files?",
+  "heading": "cyDo you want to upload files to support your counterclaim?",
+  "paragraph1": "cyWitness statements do not need to be uploaded at this stage, directions for the exchange of witness statements will be given later.",
+  "paragraph2": "cyThe claimant will be able to see these files.",
+  "options": {
+    "yes": "cyYes",
+    "no": "cyNo"
+  },
```
scottstewart-sl: You should be able to pull these from the `common` namespace

src/main/modules/steps/formBuilder/fileUploadUtils.ts
```diff
@@ -0,0 +1,70 @@
+import type { Request } from 'express';
+
+import type { DocumentStorage } from '@modules/documents/storage';
+import type { BuiltFormContent } from '@modules/steps/formBuilder/formFieldConfig.interface';
+
+/** Hidden MOJ multifile inputs post as uploadedDocuments[]; map back for error re-render. */
+export function parseUploadedDocumentsFromBody(body: Record<string, unknown>): Record<string, unknown>[] {
```
scottstewart-sl, line 14: There's already some code in `master` and also maybe a bit more in `HDPI-4270-cui-genapps-document-upload`, that handles document uploading. Does any of this new code overlap with what is there already? 

e.g. https://github.com/hmcts/pcs-frontend/blob/HDPI-4270-cui-genapps-document-upload/src/main/assets/js/multi-file-upload.ts

Otherwise this feels like a lot of new code when the file upload functionality should already be supported,

sdmiddletonSolirius: - The only duplicated logic is ~5 lines of GET-side URL wiring inlined in `formBuilder/index.ts` on HDPI-4270. On HDPI-5181 it is extracted to `fileUploadUtils.ts` as `wireFileUploadUrls`. On merge, the inline block is replaced by that call - should be a minimal localised conflict in `index.ts`.

- `wireFileUploadOnPostError` and `parseUploadedDocumentsFromBody` have no equivalent on HDPI-4270. They restore previously uploaded files when the server re-renders after a validation error, so uploads do not disappear from the UI (and the MOJ component is not left in a broken state).

- Master and HDPI-4270 have no mechanism for that rehydration. The gap is dormant on HDPI-4270 because its upload steps (`respond-to-claim/upload-document` and `make-an-application/upload-documents-to-support-your-application`) are `required: false`, so file-required validation does not fire. HDPI-5181 introduces the first mandatory-style check (validate: at least one file on counter-claim upload), which makes the POST error re-render path important. Without the new helpers, a failed submit after uploads would not wire `uploadedDocuments[]` back into `component.value`, so the MOJ list would render empty on re-render.

src/main/assets/locales/cy/respondToClaim/counterClaimDoYouWantToUploadFiles.json
sdmiddletonSolirius: Now updated :-)

src/main/modules/steps/formBuilder/fileUploadUtils.ts
paddy-hmcts, line 14: not entirely sure why this function is needed, we have `multer` which should be parsing any `multipart/form-data`. You should in theory be able to get the files by using `req.files`?

src/main/steps/utils/counterClaimUploadWanted.ts
```diff
@@ -0,0 +1,8 @@
+import type { Request } from 'express';
+
+export const counterClaimUploadWanted = (req: Request): boolean => {
+  return (
+    req.res?.locals?.validatedCase?.data?.possessionClaimResponse?.defendantResponses?.counterClaimWantToUploadFiles ===
```
paddy-hmcts: I've already created a facade/hydrated model in ccdCaseDataModel, we  should be using that to expose any properties (and not accessing `data` directly here)

paddy-hmcts (approved):
generally looks fine, a few comments

src/main/steps/utils/counterClaimUploadWanted.ts
sdmiddletonSolirius: Hi Paddy, good suggestion, I've updated accordingly and added getter to `CcdCaseModel`

src/main/modules/steps/formBuilder/fileUploadUtils.ts
sdmiddletonSolirius: When the main step form is submitted the MOJ widget posts hidden <input name="uploadedDocuments[]"> fields containing JSON metadata (filenames, ids, etc.) - that arrives in `req.body`, not `req.files`. `parseUploadedDocumentsFromBody` is only for the validation-error re-render path, so we can restore the uploaded-file list in the component after a failed POST.

scottstewart-sl (approved):

sdmiddletonSolirius:
- Removed stale counter-claim uploads from CDAM/draft when the user answers "No" on `counter-claim` or `counter-claim-do-you-want-to-upload-files`.

- Fixed deleting one pre-populated document wiping others in CCD and falsely failing validation; saves now use holistic draft updates and fresh CCD reads on upload/delete.

- Added dropzone focus highlighting when using the error summary or keyboard focus.

- Save for later on `counter-claim-upload-files` requires at least one document before returning to the dashboard.

- Confirmed post-upload routing for rent-arrears-only, non-rent-only, and mixed grounds journeys.

Screenshot 2026-06-01 at 17 37 06

PR #1305, opened 2026-05-11
HDPI-4573: View general applications

scottstewart-sl:
### Jira link

See [HDPI-4573](https://tools.hmcts.net/jira/browse/HDPI-4573)

### Change description

Display the general applications for the current user and also those from other parties that are visible to this user.

Corresponding pcs-api PR: https://github.com/hmcts/pcs-api/pull/1802

image

### Testing done

Tested locally with corresponding pcs-api changes

  * [ ] Yes
  * [x] No

### Checklist


src/main/assets/locales/cy/common.json
```diff
@@ -36,10 +36,6 @@
   },
   "change": "cyChange",
   "back": "cyBack",
-  "options": {
```
m014389b: Is options not still being used by written-terms/index.ts LOC 30/31 (among other screens)

src/main/routes/viewAllApplications.ts
```diff
@@ -0,0 +1,116 @@
...
+        };
+        partyGenAppsMap.set(partyId, partyGenApps);
+      }
+      partyGenApps.genApps.push(genAppSummary);
+
+      return partyGenAppsMap;
+    }, initialMap);
+}
+
+export default function viewAllApplicationsRoutes(app: Application): void {
```
m014389b: Preference: Could this be the 1st method in the class, makes it easier for readability

m014389b: Also I can't see any usages of this function, how is it being invoked?

src/main/assets/locales/cy/common.json
scottstewart-sl: Yes, but this is actually a duplicate of the larger `options` block on line 31 above. It was just a result of various PRs being merged with this block added in different places.

src/main/routes/viewAllApplications.ts
scottstewart-sl: There's some code that scans all the routes in this folder, so it doesn't need to be referenced explicitly:

See
https://github.com/hmcts/pcs-frontend/blob/master/src/main/app.ts#L59-L64

I don't think it actually needs a function name here, but the other routes that I checked had one, so I added one here for consistency.

scottstewart-sl: Done - I wasn't sure if the exported method should be first or last in a Typescript project.

src/main/assets/locales/cy/common.json
m014389b: D'oh, completely missed that

m014389b (approved):

src/main/middleware/caseReference.ts
```diff
@@ -32,7 +32,7 @@ export async function caseReferenceParamMiddleware(
       return next(new HTTPError('Authentication required', 401));
     }
 
-    const validatedCase = await ccdCaseService.getCaseById(accessToken, sanitisedCaseReference);
+    const validatedCase = await ccdCaseService.getCaseByIdForEvent(accessToken, sanitisedCaseReference);
```
scottstewart-sl, line 35: This has been renamed to match the same name as used in https://github.com/hmcts/pcs-frontend/pull/1276/changes#diff-d47dec0146311d7a85b189625eaeba2175c8730d6665aeae89f6c4a7d5d8e220R177

src/main/routes/documentDownload.ts
```diff
@@ -0,0 +1,43 @@
...
+import { HTTPError } from '../HttpError';
+import { oidcMiddleware } from '../middleware';
+
+import { Logger } from '@modules/logger';
+import { documentClient } from '@services/documentClient';
+
+export default function downloadDocumentRoutes(app: Application): void {
+  const logger = Logger.getLogger('document-download');
+
+  app.get('/documents/:documentId/download', oidcMiddleware, async (req: Request, res: Response) => {
```
adusumillipraveen: Not sure if there is a misalignment here, we have a download file story in citizen dashboard too and the paths is different. https://github.com/hmcts/pcs-frontend/pull/1324

scottstewart-sl: Let's chat about this directly. I based this URL on the document download PoC at https://github.com/hmcts/pcs-frontend/pull/659/changes#diff-ac34232dfef0fc7372fe496b793227a06fff7878f2c241f52b425f6d33c32455R10 and also proxied the requests through the pcs-api, (like the PoC), so that the filename manipulation only happens in one place.

The advantage of your approach is that the document ID has to be valid on the case reference as well. I'm happy with either approach.

adusumillipraveen (approved):

src/main/routes/documentDownload.ts
scottstewart-sl: As discussed previously, I replaced this code with a copy of the code from that other branch.

PR #1317, opened 2026-05-12
HDPI-5371 text fields accept cross scripting

sdmiddletonSolirius:
### Jira link

See [HDPI-5371](https://tools.hmcts.net/jira/browse/HDPI-5371)

### Change description

- Added the `xss` library https://jsxss.com/en/index.html to sanitise `text`, `textarea`, and `character-count` fields in `validateForm` via `stripHtmlTags` (FilterXSS, empty whitelist, tags stripped). This replaces the earlier denylist that blocked input with validation errors.
- HTML is removed without showing an error so attackers are not told their payload was detected. 
- We only run xss when input looks like HTML (`looksLikeHtml`), so legitimate text such as `1 bedroom is < 2 bedrooms` is not truncated or stored as entities.
- Moved `stripHtmlTags` call to before the `isMissing` check in `validateForm` so that input consisting solely of stripped tags (e.g. '<script>...</script>') correctly fails required field validation rather than being silently accepted.

**BEFORE**
Screenshot 2026-06-01 at 10 31 45

**AFTER**
Screenshot 2026-06-01 at 10 31 54

  * [ ] Yes
  * [X] No

### Checklist


src/main/steps/utils/fieldValidators.ts
```diff
@@ -1,5 +1,18 @@
 export const EMOJI_PATTERN = /\p{Emoji_Presentation}|\p{Extended_Pictographic}|\u200D|\uFE0F/u;
 
+export const UNSAFE_TEXT_PATTERNS: RegExp[] = [
```
paddy-hmcts: I think perhaps it would be much better to use an npm library here that's been tried and tested:

https://www.npmjs.com/package/xss

sdmiddletonSolirius: Hi @paddy-hmcts - Have utilisted the xss package as discussed, with some modifications to allow legitimate use of the comparison operators.

src/main/assets/locales/cy/common.json
```diff
@@ -48,6 +48,7 @@
     "defaultRequired": "cyThis field is required",
     "defaultInvalid": "cyInvalid format",
     "defaultMaxLength": "cyMust be {max} characters or fewer",
+    "defaultSpecialCharacter": "{fieldName} must only include letters a to z, and special characters such as hyphens, spaces and apostrophes",
```
andrewdami11, line 53: Sibling Welsh keys use a `cy` prefix; this new `defaultSpecialCharacter` value is English. Should this use Welsh copy (or the same `cy…` placeholder pattern) for consistency?

src/test/unit/modules/steps/formBuilder/helpers.test.ts
```diff
@@ -1517,6 +1517,105 @@ describe('formBuilder helpers', () => {
...
+                },
+              },
+              { value: 'notSure' },
+            ],
+          },
+        ];
+
+        const errors = validateForm(req, fields, {});
+
+        expect(errors['tenancyTypeConfirm.correctType']).toBeUndefined();
```
andrewdami11, line 1583: Could we add a test where a required field only contains _<script>…</script>_? After stripping it should be empty, and validation should still fail, no?

src/main/assets/locales/cy/common.json
sdmiddletonSolirius: Hey Andrew - the cy prefix is already built into the fieldName, so it's removed here as it was generating a double cycy on render - I learned this the hard way first time round.

src/test/unit/modules/steps/formBuilder/helpers.test.ts
sdmiddletonSolirius: Good catch - this has now been updated so inputs with only script tags will fail validation for being an empty response, before they were silently failing.

andrewdami11 (approved):

src/main/modules/steps/formBuilder/helpers.ts
```diff
@@ -470,28 +486,31 @@ export function validateForm(
...
+          }
+
+          const translationLabelKey =
+            typeof field.translationKey === 'object' ? field.translationKey.label : field.translationKey;
+
+          const resolvedLabel = translationLabelKey && t ? getTranslation(t, translationLabelKey) : undefined;
+
+          const displayName = resolvedLabel ?? toSentenceCase(fieldName);
+
+          const defaultSpecialCharacterMsg = translations?.defaultSpecialCharacter?.replace('{fieldName}', displayName);
```
paddy-hmcts: we shouldn't be doing this with translations, but using interpolation instead: https://www.i18next.com/translation-function/interpolation

src/main/modules/steps/formBuilder/helpers.ts
```diff
@@ -470,28 +486,31 @@ export function validateForm(
...
+          errors[fieldName] =
+            defaultSpecialCharacterMsg ||
+            `${displayName} must only include letters a to z, and special characters such as hyphens, spaces and apostrophes`;
```
paddy-hmcts: I don't think we should be providing default fallbacks for translation strings as the dev should know where it's broken and fix appropriately

src/main/modules/steps/formBuilder/helpers.ts
```diff
@@ -470,28 +486,31 @@ export function validateForm(
...
-
-              const defaultSpecialCharacterMsg = translations?.defaultSpecialCharacter?.replace(
-                '{fieldName}',
-                displayName
-              );
-              errors[fieldName] =
-                defaultSpecialCharacterMsg ||
-                `${displayName} must only include letters a to z, and special characters such as hyphens, spaces and apostrophes`;
-            }
+            setDefaultSpecialCharacterError();
```
paddy-hmcts: can I ask why this is actually required here? Isn't this part of a separate ticket that was already addressed? more for my own understanding

src/test/unit/app/controller/validation.test.ts
```diff
@@ -104,5 +104,49 @@ describe('validateForm', () => {
       expect(errors.textareaField).toContain('must only include letters a to z');
       expect(errors.charCountField).toContain('must only include letters a to z');
     });
+
```
paddy-hmcts, line 107: can we please remove this before merging

paddy-hmcts (approved):
few minor comments but overall looks good

src/main/modules/steps/formBuilder/helpers.ts
sdmiddletonSolirius: Hi @paddy-hmcts - you are correct, this was originally part of the initial approach which was to reject illegal characters, and so we were piggy backing off the same error validation for emoji's - however, since adopting the xss sanitisation approach, I've now reverted this back so it doesn't encroach on the emoji implementation which is part of a separate ticket.

src/test/unit/app/controller/validation.test.ts
sdmiddletonSolirius: Done

src/main/modules/steps/formBuilder/helpers.ts
sdmiddletonSolirius: Have now reverted to before XSS implementation

src/main/assets/locales/cy/common.json
paddy-hmcts: you probably also no longer need this key now

paddy-hmcts (approved):

PR #1324, opened 2026-05-12
File download in view-documents

adusumillipraveen:
### Jira link

See https://tools.hmcts.net/jira/browse/HDPI-6425 

### Change description

- Streams document from CDAM directly and renders in the browser where needed and downloads if browser cannot render.
- Also uses the document uuid in url ( not the actual dm-store id )
- Matches that the document is present in the alldocument list which user has access to.

### Testing done

Manually demoed and tested it all works

  * [ ] Yes
  * [ ] No

### Checklist


src/main/routes/viewDocuments.ts
```diff
@@ -4,51 +4,106 @@ import { HTTPError } from '../HttpError';
...
+        const stream = binaryResponse.stream;
+        const contentType = asHeaderString(binaryResponse.contentType);
+        const contentLength = asHeaderString(binaryResponse.contentLength);
+        const contentDisposition = asHeaderString(binaryResponse.contentDisposition);
+
+        res.setHeader('Content-Type', contentType || 'application/octet-stream');
+        if (contentLength) {
+          res.setHeader('Content-Length', contentLength);
+        }
+        res.setHeader('Content-Disposition', contentDisposition || `inline; filename="${toSafeFilename(filename)}"`);
```
scottstewart-sl: According to [MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Content-Disposition#:~:text=It%27s%20recommended%20to%20include%20both%20for%20maximum%20compatibility) it's recommended to include both `filename` and `filename*`. Also just removing the double quotes might not be enough to make it suitable for the `filename` attribute.

The code that I had previously taken from the (Java) PoC for document download had this sanitising RegEx, `filename.replaceAll("[\\\\/:*?\"<>|]", "_")` but even that wouldn't handle some non-ASCII chars like é

src/main/utils/httpHeaders.ts
```diff
@@ -0,0 +1,13 @@
+export function asHeaderString(value: unknown): string | undefined {
+  if (typeof value === 'string') {
+    const trimmed = value.trim();
+    return trimmed || undefined;
+  }
+  if (typeof value === 'number') {
+    return String(value);
+  }
+  if (Array.isArray(value) && value.length > 0) {
+    return String(value[0]);
```
scottstewart-sl: Minor - if you're trimming the `string` types, should you also trim the first value from an array?

scottstewart-sl (approved):

fudge88 (approved):

PR #1413, opened 2026-05-26
HDPI-4181-error-handling-screens

AlexBuzea:
### Jira link

See [HDPI-4181](https://tools.hmcts.net/jira/browse/HDPI-4181)

### Change description

-refactoring `error.njk` to cover different content based on the status codes
-added temporary route for testing in dev/preview environment which allows passing the `status code` and `Retry-After` header by allowing to pass query parameters to the route and mock different responses (to be removed before merging, for manual testing purposes only)
-for 502, 503, 504, 429 service unavailable statuses it renders 3 different paragraphs depending on whether `Retry-After` header is provided: a fallback to try again later if not provided, if it returns seconds it converts them to minutes and if it provides a date and time it's added directly in the paragraph

image
image
image

Service unavailable variations:
- image
- image
- image

### Testing done

-tested manually and added unit tests

  * [ ] Yes
  * [x] No

### Checklist


src/main/routes/devErrorPreview.ts
```diff
@@ -0,0 +1,12 @@
+import type { Application, NextFunction, Request, Response } from 'express';
+
+import { HTTPError } from '../HttpError';
+
+// Temporary route to preview error pages in preview environment. Not intended for production use.
+export default function (app: Application): void {
+  app.get('/dev/error/:status', (req: Request, _res: Response, next: NextFunction) => {
```
andrewdami11: Is this temporary preview route actually only gated to the preview env? It seems like it's currently registered in all environments, which may expose a test-only endpoint in production.

src/main/routes/devErrorPreview.ts
```diff
@@ -0,0 +1,12 @@
...
+    const statusCode = Number(req.params.status);
```
andrewdami11: Would it be worth validating `status` so this preview route only accepts the error codes listed in the ticket?
At the moment it can return any status, which could lead to confusing or unintended responses, while this endpoint is meant to simulate only the agreed error scenarios.

src/main/routes/devErrorPreview.ts
AlexBuzea: @andrewdami11 It is not, I had a guard in this commit but it wouldn't show on preview either https://github.com/hmcts/pcs-frontend/pull/1413/changes/d3d01090f95c025e4f474d3c85a9d80d22f637c1 . So my thinking here is to remove the route before merging as this was added strictly for testing the UI

src/main/routes/devErrorPreview.ts
AlexBuzea: @andrewdami11 You could but this code is intentionally very minimal for testing and it's not intended to go into production

andrewdami11 (approved):

src/main/routes/devErrorPreview.ts
```diff
@@ -0,0 +1,12 @@
+import type { Application, NextFunction, Request, Response } from 'express';
+
+import { HTTPError } from '../HttpError';
+
+// Temporary route to preview error pages in preview environment. Not intended for production use.
```
AlexBuzea: To be removed before merging to master

src/test/unit/routes/devErrorPreview.test.ts
```diff
@@ -0,0 +1,14 @@
+import type { Application } from 'express';
+
+import devErrorPreview from '../../../main/routes/devErrorPreview';
+
+//Temporary tests to ensure the route is registered correctly for testing error pages in the preview environment. Not intended for production use.
```
AlexBuzea: To be removed before merging to master

libanAbdirahman1 (approved):

PR #1494, opened 2026-06-09
hotfix casenumber

toby-plunkett:
Hotfix to adjust tests in preview environment where case type no longer contains the PR number at the end.

### DIFF
--- a/src/test/ui/data/api-data/createCase.api.data.ts
+++ b/src/test/ui/data/api-data/createCase.api.data.ts
@@ -13,7 +13,5 @@ export const createCaseApiData = {
     },
     legislativeCountry: 'England',
   },
-  createCaseApiEndPoint: `/case-types/PCS${
-    process.env.PCS_API_CHANGE_ID ? '-' + process.env.PCS_API_CHANGE_ID : ''
-  }/cases`,
+  createCaseApiEndPoint: `/case-types/${process.env.CASE_TYPE_ID ?? 'PCS'}/cases`,
 };
--- a/src/test/ui/data/api-data/createCaseEventToken.api.data.ts
+++ b/src/test/ui/data/api-data/createCaseEventToken.api.data.ts
@@ -12,7 +12,5 @@ export const createCaseEventTokenApiData = {
     },
   }),
 
-  createCaseEventTokenApiEndPoint: `/case-types/PCS${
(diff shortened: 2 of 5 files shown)


libanAbdirahman1 (approved):

src/test/ui/data/api-data/createCase.api.data.ts
```diff
@@ -13,7 +13,5 @@ export const createCaseApiData = {
     },
     legislativeCountry: 'England',
   },
-  createCaseApiEndPoint: `/case-types/PCS${
-    process.env.PCS_API_CHANGE_ID ? '-' + process.env.PCS_API_CHANGE_ID : ''
-  }/cases`,
+  createCaseApiEndPoint: `/case-types/${process.env.CASE_TYPE_ID ?? 'PCS'}/cases`,
```
benouaer, line 16: why not just set it to PCS?

src/test/ui/data/api-data/createCaseEventToken.api.data.ts
```diff
@@ -12,7 +12,5 @@ export const createCaseEventTokenApiData = {
     },
   }),
 
-  createCaseEventTokenApiEndPoint: `/case-types/PCS${
-    process.env.PCS_API_CHANGE_ID ? '-' + process.env.PCS_API_CHANGE_ID : ''
-  }/event-triggers/createPossessionClaim`,
+  createCaseEventTokenApiEndPoint: `/case-types/${process.env.CASE_TYPE_ID ?? 'PCS'}/event-triggers/createPossessionClaim`,
```
benouaer, line 15: is it going to be anything other than PCS?

benouaer (approved):

src/test/ui/data/api-data/createCase.api.data.ts
toby-plunkett: it is changed to PCS-Staging at some points during deployment. I'm unsure whether this makes a difference to this specific endpoint, but leaving this functionality in does allow us to change it in the future via env vars if we need

src/test/ui/data/api-data/createCaseEventToken.api.data.ts
toby-plunkett: same as above

PR #1567, opened 2026-06-18
new role access for citizen and caseworker-pcs-solicitor role

fudge88:
### Jira link

See [HDPI-7211](https://tools.hmcts.net/jira/browse/HDPI-7211)

### Change description

Introduced access control on CUI for:
- citizen role
- caseworker-pcs-solicitor role.

| URL | citizen | caseworker-pcs-solicitor | any other / no role | 
|---|---|---|---|
| `/claims` | allow | 403 | 403 | 
| `/access-your-case` | allow | 403 | 403 | 
| `/case/:ref/dashboard` | allow | 403 | 403 | 
| `/case/:ref/make-an-application` | allow | 403 | 403 | 
| `/case/:ref/respond-to-claim/...` | allow | allow | 403 | 

### Citizen:

/claims:
image

/access-your-case:
image

/case/:ref/dashboard:
image

/case/:ref/make-an-application:
image

/case/:ref/respond-to-claim:
image

/case/:ref/upload-additional-documents
image

### Caseworker-pcs-solicitor:

/claims:
image

/access-your-case:
image

/case/:ref/dashboard:
image

/case/:ref/make-an-application:
image

/case/:ref/respond-to-claim:
image

/case/:ref/upload-additional-documents
image

  * [ ] Yes
  * [x] No

### Checklist


src/main/steps/index.ts
```diff
@@ -67,6 +71,7 @@ export const journeyRegistry: Record<string, JourneyConfig> = {
       flowConfig: makeAnApplicationFlowConfig,
       stepRegistry: makeAnApplicationStepRegistry,
     },
+    requiredRoles: CITIZEN_USER_ROLES,
   },
   uploadAdditionalDocuments: {
```
andrewdami11, line 78: The `uploadAdditionalDocuments` journey has no `requiredRoles` and no matching `accessRules` entry, so `caseworker-pcs-solicitor` users can still reach `/case/:ref/upload-additional-documents`. Should that journey be citizen-only like dashboard and make-an-application?

src/main/modules/oidc/oidc.ts
```diff
@@ -206,6 +209,25 @@ export class OIDCModule {
...
+            path: req.session.returnTo ?? '',
+            userId: typeof user.uid === 'string' ? user.uid : undefined,
+            userRoles,
+            rule: accessDecision.rule,
+          });
+
+          delete req.session.codeVerifier;
+          delete req.session.nonce;
+          delete req.session.returnTo;
+          req.session.save(() => res.redirect('/login'));
```
andrewdami11, line 227: When login-time access is denied, the user is sent back to `/login` with `returnTo` cleared and no on-screen explanation. Would an error message (or something along those lines) help users understand why sign-in did not complete?

src/main/routes/registerSteps.ts
```diff
@@ -159,6 +160,34 @@ export function registerSteps(router: IRouter, specificJourney?: string): void {
   });
 }
 
+/**
+ * Verify the journey's runtime requireRoles guard agrees with the URL-pattern
+ * rule used by the OAuth callback gate. If they ever drift, fail at boot.
+ */
+function assertRoleRuleAlignment(
```
andrewdami11, line 186: `assertRoleRuleAlignment` only runs for registry journeys with `requiredRoles`, not for standalone routes like `/claims` and `/access-your-case`. Would extending the boot-time drift check to those routes help keep `accessRules` and route guards in sync?

src/main/app.ts
```diff
@@ -47,7 +43,7 @@ app.use((req, res, next) => {
 });
 
 app.use(pageTrackingUrlMiddleware);
-app.use(legalRepresentativeAccessMiddleware);
+app.use(roleAccessMiddleware);
```
andrewdami11, line 46: `roleAccessMiddleware` already enforces `accessRules` globally, and several routes/journeys also mount `requireRoles` for the same paths. Would it be worth consolidating to a single enforcement layer to reduce duplication?

src/main/modules/oidc/oidc.ts
andrewdami11, line 243: After a successful OIDC callback with no `returnTo`, users are redirected to `/claims`, which is citizen-only and will return 403 for solicitors. Would a role-aware default landing URL be worth considering?

src/main/modules/oidc/oidc.ts
fudge88: this has been implemented in civil - business has not given a requirement for this - so its agreed to have this at current.

src/main/modules/oidc/oidc.ts
fudge88: solicitors login via XUI where they will be redirected to CUI's respond-to-claim. 
They wont be logging in directly from CUI. This was added as a safety precaution.

src/main/steps/index.ts
fudge88: I've added requiredRoles to the journey and a matching upload-additional-documents rule in accessRules.ts, so it's now citizen-only like the dashboard and make-an-application. A solicitor gets a 403 on that path now, and since the journey declares requiredRoles, the boot-time check would fail if the rule and the journey ever fall out of sync.

this has also been added to the AC's for the ticket

src/main/routes/registerSteps.ts
fudge88: Rather than add another assertion, I pointed the standalone guards (/claims, /access-your-case, dashboard) straight at accessRules - they now pull their roles via rolesForRule('claims') etc. instead of restating CITIZEN_USER_ROLES. That means the guards and accessRules can't drift apart by construction, and a bad/renamed rule name throws at boot. I went a step further and made the journey requiredRoles derive the same way, so accessRules is now the single source of truth for every gated route.

src/main/app.ts
fudge88: The two layers are intentional: the global middleware is the backstop and the per-route/per-journey guards give locality, so it's defence-in-depth rather than accidental duplication. The drifts now tightened up: everything derives from accessRules, so the same rule can't be authored in two places anymore.

src/main/routes/registerSteps.ts
```diff
@@ -187,6 +216,21 @@ export function registerAllJourneys(app: Application): void {
     // Apply journey-specific middleware
     // Note: Auto-save is handled via formBuilder's beforeRedirect, not middleware
     journeyRouter.param('caseReference', caseReferenceParamMiddleware);
+
+    // Authenticate before any role or case-access guard. requireRoles skips
+    // unauthenticated users and requireEventAccess 401s them, so without this an
+    // unauthenticated request would hit a 401/403 instead of being redirected to login.
+    journeyRouter.use(basePath, oidcMiddleware);
```
andrewdami11, line 242: Now that `oidcMiddleware` runs at journey `basePath`, is per-step `oidcMiddleware` still needed, or could it be removed to avoid running auth twice?

fudge88: You're right - thanks - i missed this completely. Ive removed the double oidc call

andrewdami11 (approved):

sdmiddletonSolirius (approved):
LGTM

PR #1643, opened 2026-07-01
HDPI-7401 - Add additional defendants DOB and Addresses in two pages

mazzopardi2:
### Jira link

See [HDPI-7401](https://tools.hmcts.net/jira/browse/HDPI-7401)

### Change description

- toPartialParty() now returns address and date of birth for other parties.

### Testing done

- View the claim: show other defendants’ and underlessees’ addresses and show "Address unknown" and "Persons unknown"  when fields blank
- View the response: show other defendants’ address and DOB when available and hide DOB when address is unknown 
- Renamed shared locale keys

  * [ ] Yes
  * [x] No

### Checklist


arun-hmcts (approved):

src/main/routes/viewTheResponse.ts
```diff
@@ -150,12 +144,33 @@ function joinName(firstName?: string, lastName?: string): string {
...
+  t: TFunction,
+  party: CcdDefendantParty | CcdParty | undefined,
+  propertyAddress?: CcdCaseAddress
+): string {
+  if (isNo(party?.addressKnown)) {
+    return t('viewTheResponse:addressUnknown');
+  }
+
+  if (isYes(party?.addressSameAsProperty) && propertyAddress) {
+    return formatAddress(propertyAddress) ?? '';
```
scottstewart-sl: Minor - In theory `propertyAddress` should always be present shouldn't it? So maybe this condition could be simpler and not check for `propertyAddress`, and return `t('viewTheResponse:addressUnknown')` in the unlikely event that it isn't defined?

Up to you. 🙂

src/main/services/ccdCase.interface.ts
```diff
@@ -139,6 +139,11 @@ export interface CcdParty {
   firstName?: string;
   lastName?: string;
   orgName?: string;
+  nameKnown?: string;
+  address?: CcdCaseAddress | Record<string, never>;
```
scottstewart-sl: Just out of interest, why does this need two types? Wouldn't the data always fit into `CcdCaseAddress` ?

scottstewart-sl (approved):

src/main/services/ccdCase.interface.ts
mazzopardi2: To be honest, I followed the pattern in CCDDefendantParty and CcdClaimantEnteredDefendantDetails in the same file without thinking about it but you're right the data should fit in CCDCaseAddress like you said. I've updated - good point! :D

src/main/routes/viewTheResponse.ts
mazzopardi2: Good catch, I've updated.

PR #1676, opened 2026-07-07
Extended e2e frameowrk to support running the e2e tests in Welsh language

Gautham059:
### Jira link

https://tools.hmcts.net/jira/browse/HDPI-7406

### Change description

1. Create index.selector.ts file for seleting the data files based upon an environment variable.
2. Create a data file under new folder page-data-cy. Ex: src/test/ui/data/page-data-cy/dashboard.page.data.cy.ts and create an index.ts file with export list.

### Testing done
Data file changed when env variable is set 
image


Jenkinsfile_CNP
SameenaHMCTS, line 1: we might need similar options/flags in jenkin nightly also to support welsh execution in nightly?

srinijg (approved):

Jenkinsfile_CNP
Gautham059: As mentioned, there will be another discussion for nightly runs as running these tests may overload the pipeline. When we decide to run selected tests, I'll create another PR with fully controlled config

SameenaHMCTS (approved):
lgtm

scottstewart-sl (approved):

PR #1757, opened 2026-07-21
CUI Your Support RA

toby-plunkett:
### Jira link

See [HDPI-7379](https://tools.hmcts.net/jira/browse/HDPI-7379)

### Change description
This PR adds our integration with CUI Your Support - Reasonable Adjustments. 

**In Summary:** 
Frontend: [here](https://github.com/hmcts/pcs-frontend/pull/1757/)
- Allows us to navigate to Your Support Microsite via Check your answers on Respond to claim. 
- Retrieves the Adjustments requested (as flags)
- Posts them to Draft Database after submission & at end of Respond to claim journey. 
 
Backend: [here](https://github.com/hmcts/pcs-api/pull/2284)
- Allows Reasonable adjustments to be saved and sent as draft data. 
- Persists Reasonable adjustments as case flags. 
- Adds Unique constraint to flag_ref_data to limit potential for exploitation. 

### Testing done
Many unit tests have been added. 
As well as manual testing. A video will be also provided.

  * [ ] Yes
  * [x] No

### Checklist


src/main/steps/respond-to-claim/reasonable-adjustments-cancelled/index.ts
```diff
@@ -0,0 +1,22 @@
+import type { Request } from 'express';
```
toby-plunkett, line 1: Placeholder until we get confirmation from service what text they'd like.

src/main/steps/respond-to-claim/sections.config.ts
```diff
@@ -137,14 +137,16 @@ const sectionDefs = [
       'incomeAndExpenditure',
       'uploadFiles',
     ],
-    // 'reasonable-adjustments-triage', 'equality-and-diversity-start' and
+
+
```
toby-plunkett: looks messy but just reworded so it makes sense now that RA is implemented

src/main/routes/reasonableAdjustmentsCallback.ts
```diff
@@ -0,0 +1,87 @@
...
+      if (!serviceToken) {
+        logger.error(`No S2S service token available to fetch Your Support payload for id ${payloadId}`);
+        return safeRedirect303(res, errorUrl, fallback, ['/case']);
+      }
+
+      try {
+        const payload = await cuiRaService.getPayload(payloadId, serviceToken);
+        logger.info(
+          `Your Support payload received for case ${caseReference}, id ${payloadId}, action ${payload.action}`
+        );
```
gergelykisshmcts, line 64: payloadId comes straight from the URL with nothing tying it to this case or user + S2S fetch happens before any case-access check so a logged-in user can pull another party's adjust. flags into their own draft. paymentReturn.ts:20 and counterClaimPaymentStart.ts:28 both bail out unless the matching reference is in the session so I think we need something like: payload.correlationId === caseReference at least or check what happens there and do something like that.

src/main/routes/reasonableAdjustmentsCallback.ts
```diff
@@ -0,0 +1,87 @@
+import config from 'config';
```
gergelykisshmcts: Unlike every journey page it sits outside respondToClaimFeatureMiddleware: the draft write stays reachable even when the respond-to-claim flag is off.

src/test/unit/services/cuiRa/startYourSupport.test.ts
```diff
@@ -0,0 +1,180 @@
...
+      accessToken: 'idam-access-token',
+      serviceToken: 's2s-token-value',
+      body: {
+        callbackUrl: 'http://frontend/case/1234123412341234/respond-to-claim/reasonable-adjustments/callback/:id',
+        logoutUrl: 'http://frontend/logout',
+        language: 'en',
+        existingFlags: { partyName: 'John Doe', roleOnCase: 'Defendant', details: [] },
+        hmctsServiceId: 'AAA3',
+        masterFlagCode: 'RA0001',
+        correlationId: '1234123412341234',
```
gergelykisshmcts, line 81: This looks to me (correct me if i'm wrong): using caseReference as the correlationId so wont be unique per invocation, so there's nothing to validate the retur against. 
Check out pcqRedirect.ts: uses uuid() per invocation and persist it before redirecting out.

src/main/steps/respond-to-claim/reasonable-adjustments-triage/index.ts
```diff
@@ -1,14 +1,39 @@
...
+      req.res?.redirect(303, redirectUrl); // postHandler short-circuits on res.headersSent
+    } catch (error) {
+      // Any failure launching Your Support (cui-ra down, POST error, missing token) lands the
+      // citizen on the context-aware RA error page
+      logger.error(`Failed to launch Your Support for case ${caseReference}`, error);
+      if (caseReference) {
+        req.res?.redirect(303, `/case/${caseReference}/respond-to-claim/reasonable-adjustments-error`);
+      }
+    }
+  },
```
gergelykisshmcts, line 41: When caseReference is missing this catch neither redirects nor rethrows, so falls through and the user silently continues as though they'd skipped the questions. Is that intentional?

gergelykisshmcts (changes requested):

src/main/routes/reasonableAdjustmentsCallback.ts
toby-plunkett: Addressed - binding payload correlation id to case reference

src/main/steps/respond-to-claim/reasonable-adjustments-triage/index.ts
toby-plunkett: Addressed - Throws to shared error handler

src/main/routes/reasonableAdjustmentsCallback.ts
toby-plunkett: addressed in next commit! :)

src/test/unit/services/cuiRa/startYourSupport.test.ts
toby-plunkett: Very good spot, I had this in my implementation BEFORE I had the correlationId being used. (as it's in the payload spec). CorrelationID matches the case reference in this test now.

gergelykisshmcts (approved):

charts/pcs-frontend/values.yaml
```diff
@@ -40,4 +40,10 @@ nodejs:
     FEE_SERVICE_URL: 'http://fees-register-api-{{ .Values.global.environment }}.service.core-compute-{{ .Values.global.environment }}.internal'
     REDIRECTS_MANAGE_CASE_RETURN_URL: 'https://manage-case.{{ .Values.global.environment }}.platform.hmcts.net/cases/case-details/PCS/PCS'
     PAYMENT_RETURN_URL: 'https://pcs.{{ .Values.global.environment }}.platform.hmcts.net/payment/return'
+    # TODO: prod uses a different cui-ra host (manage-your-support-for-hmcts-services.service.gov.uk),
+    # not the cui-ra.<env>.platform.hmcts.net pattern below — prod needs an explicit override so the
+    # invoke/callback URLs resolve correctly.
+    CUI_RA_URL: 'https://cui-ra.{{ .Values.global.environment }}.platform.hmcts.net'
+    CUI_RA_CALLBACK_URL: 'https://pcs.{{ .Values.global.environment }}.platform.hmcts.net/case/:caseReference/respond-to-claim/reasonable-adjustments/callback/:id'
```
andrewdami11, line 46: These callback and logout URLs use `pcs.{env}`, and `values.preview.template.yaml` does not override them the way `PAYMENT_RETURN_URL` does with `${SERVICE_FQDN}`. 
After Your Support, preview users may be sent to AAT PCS instead of this PR. Is it worth adding the same `${SERVICE_FQDN}` overrides in the preview chart, and pointing `CUI_RA_URL` at a real cui-ra host if `cui-ra.preview` does not exist?

src/main/routes/reasonableAdjustmentsCallback.ts
```diff
@@ -0,0 +1,108 @@
...
+        // (or any other value) means nothing was sent → the "No request was sent" page.
+        if (payload.action !== 'submit') {
+          return safeRedirect303(res, cancelledUrl, fallback, ['/case']);
+        }
+
+        // Persist the returned flags to the case DRAFT via the same citizen respondPossessionClaim
+        // draft-save our journey pages use. The draft-save fully REPLACES the defendant response (and
+        // the final submit reads this same draft), so re-send the existing answers — narrowed to the
+        // defendant slice — alongside the flags, or a flags-only post would wipe them.
+        const rawFlags = payload.replacementFlags ?? payload.flagsAsSupplied;
```
andrewdami11: On `submit`, if `replacementFlags` is missing this falls back to `flagsAsSupplied`, and if both are missing it still goes to the confirmation page. The citizen may see that a request was sent when nothing was saved, or old flags may be written back if cui-ra omits `replacementFlags`. Is it worth persisting only `replacementFlags` on submit (including empty `details`), and sending the user to the error or cancelled page when there is nothing to save?

src/main/assets/locales/en/respondToClaim/reasonableAdjustmentsError.json
```diff
@@ -0,0 +1,6 @@
+{
+  "pageTitle": "Sorry, there is a problem with the service",
+  "heading": "Sorry, there is a problem with the service",
+  "paragraph": "There was a problem starting the support questions. Try again.",
```
andrewdami11: This copy says there was a problem starting the support questions, but the same page is also used when the return callback fails (S2S, token, payload, correlation, or draft save). After a failed save the user may think Your Support never started. Is it worth widening the EN and CY copy so it covers both launch and return failures?

toby-plunkett: Adjusted wording to make sense in both scenarios. 
'There was a problem with the support questions. Try again.'

src/main/routes/reasonableAdjustmentsCallback.ts
toby-plunkett: Good catch! I've adjusted it slightly, so it only persists the replacement. Otherwise, if there are no replacement flags, we hit the 'no request sent' page.

charts/pcs-frontend/values.yaml
toby-plunkett: Updated Preview chart to used SERVICE_FQDN in url. I've left the CUI_RA_URL as CUIYS team have said they have the service running in each environment, but will keep an eye on this as it gets deployed before approving.

src/main/steps/respond-to-claim/reasonable-adjustments-triage/index.ts
```diff
@@ -1,14 +1,44 @@
...
   stepDir: __dirname,
   flowConfig,
   customTemplate: `${__dirname}/reasonableAdjustmentsTriage.njk`,
+  // "Continue to the questions" (reasonableAdjustmentsChoice=questions) launches the Your Support
+  // microsite;
+  beforeRedirect: async (req: Request) => {
+    if (req.body.reasonableAdjustmentsChoice !== 'questions') {
+      return; // "skip": let the normal next-step flow continue to language-used
+    }
+    if (!(await isCuiYourSupportEnabled(req))) {
```
andrewdami11, line 25: When the flag is off, this page still shows Continue to the questions, and a click just continues as skip.
The microsite is not launched, but the user still sees the offer. Is it worth hiding that button, or something along those lines, so the journey does not advertise _Your Support_, when it should be switched off?

src/main/middleware/cuiYourSupportFeatureMiddleware.ts
```diff
@@ -0,0 +1,17 @@
...
+// Feature gate for the CUI Your Support (Reasonable Adjustments) routes. 
+export const cuiYourSupportFeatureMiddleware: RequestHandler = async (
+  req: Request,
+  res: Response,
+  next: NextFunction
+): Promise<void> => {
+  if (await isCuiYourSupportEnabled(req)) {
+    return next();
+  }
+  return handleRespondToClaimDisabled(req, res);
```
andrewdami11: If the flag is turned off while the citizen is on the microsite, this sends them to the dashboard via `handleRespondToClaimDisabled`. Flags are not saved, which looks right for a kill switch, but they also leave respond-to-claim instead of continuing to `language-used` or cancelled, which is what triage does when the flag is off. Is it worth keeping them in the journey here too?

toby-plunkett: now moves to language used (next page)

src/main/steps/respond-to-claim/reasonable-adjustments-triage/index.ts
toby-plunkett: Went with hiding the button. I did consider skipping the Triage page entirely but thats a more major writeup and moves responsibility of the first visible step to language used page.

andrewdami11 (approved):

PR #1769, opened 2026-07-23
HDPI-7643: Number defendants according to rank for View the Claim and View the Response pages

mazzopardi2:
### Jira link

See [HDPI-7643](https://tools.hmcts.net/jira/browse/HDPI-7643)

### Change description
- numbers defendants on view-the-claim and view-the-response using real rank coming via api.
- remove additional defendant sections so defendant can only see their own details

### Testing done
- manually reviewed

  * [ ] Yes
  * [x] No

### Checklist


src/main/routes/viewTheResponse.ts
```diff
@@ -144,16 +144,10 @@ function joinName(firstName?: string, lastName?: string): string {
...
-
-  if (party.firstName === 'Person unknown' && party.lastName === 'Person unknown') {
-    return t('viewTheResponse:personsUnknown');
-  }
-
-  return joinName(party.firstName, party.lastName) || t('viewTheResponse:personsUnknown');
+function resolveCurrentDefendantRank(caseData: CcdCaseData): number {
+  const currentDefendantPartyId = caseData.possessionClaimResponse?.currentDefendantPartyId;
+  const currentDefendant = (caseData.allDefendants ?? []).find(defendant => defendant.id === currentDefendantPartyId);
+  return typeof currentDefendant?.value.rank === 'number' ? currentDefendant.value.rank : 1;
```
scottstewart-sl: I'm not sure about defaulting to 1 if the party or rank isn't found or the rank isn't a number. Could this return undefined for that error scenario and the page title just show a more generic title?

mazzopardi2: Thanks Scott, agreed defaulting to 1 could be misleading if the lookup fails so it's returning undefined now. 
I've left the section title as is so that it would surface the bug if encountered, or do you think I should also add a fallback key to have it degrade gracefully?

scottstewart-sl (approved):

src/main/routes/viewTheResponse.ts
scottstewart-sl: Based on our discussion on that chat, that approach is fine with me

src/main/utils/viewTheClaim/viewTheClaimSections.ts
```diff
@@ -85,6 +85,7 @@ export function buildDefendantSection(
   copy: ViewTheClaimCopy
 ): ViewTheClaimSection | undefined {
   const defendant = firstDefendantParty(data);
+  const rank = typeof defendant?.rank === 'number' ? defendant.rank : 1;
```
andrewdami11: On view-the-response we avoided defaulting a missing rank to 1 because it can mislabel the defendant.
Here `buildDefendantSection` still defaults to 1 (and additional defendants to `index + 2` on line 119). Should claim titles use the same undefined/no-number approach when rank is absent, so we do not show a misleading defendant number?

src/test/unit/utils/viewTheClaimUtils.test.ts
```diff
@@ -187,16 +185,16 @@ describe('viewTheClaimUtils', () => {
     expect(page.sections.map(section => section.title).slice(0, 4)).toEqual([
       'Claimant details',
       'Defendant 1 details',
```
andrewdami11, line 193: The tests only check the fallback titles (Defendant 1 / Defendant 2) and never set `rank`. 
Is it worth adding a case with ranks out of order (e.g. 2 then 1) and check the section titles match those ranks?

src/test/unit/routes/viewTheResponse.test.ts
```diff
@@ -355,30 +355,14 @@ describe('viewTheResponse route', () => {
         }),
       ])
     );
-    expect(renderArgs.defendant1Details.rows.length).toBeGreaterThan(0);
-    expect(renderArgs.defendant1Details.rows.map((row: { key: { text: string } }) => row.key.text)).toEqual([
+    expect(renderArgs.defendantDetails.rows.length).toBeGreaterThan(0);
```
andrewdami11: This updates the render key to `defendantDetails` and checks `additionalDefendantDetails` is undefined, but never asserts `defendantDetails.sectionTitle` with a real rank from `allDefendants`. 
Is it worth adding something like a matched `currentDefendantPartyId` with `rank: 2` renders the numbered defendant title?

andrewdami11 (commented):
Should this ticket be gated behind feature flags?

matthewsearle01 (approved):

PR #1832, opened 2026-08-03
HDPI-7545

ayub-macalim:
### HDPI-7545

See [HDPI-7545](https://tools.hmcts.net/jira/browse/HDPI-7545)

### Change description

On the "Contact by text message" page, when a user selected Yes they weren't asked for a phone number, so they could finish thinking they'd get text updates when we had no mobile number to text them on.
Now, selecting Yes shows a box to enter a UK mobile number, and it checks that it's actually a mobile (not a landline). The number is saved, shows again if they come back to the page, and clears if they change their answer to No. Added Welsh translations for the new text too.

### Testing done

image

image

image

- [ ] Yes
- [ ] No

### Feature flag assessment

**Feature flag:** Does this PR introduce or change user-facing behaviour that should be behind feature flag(s)?

- [ ] Yes — flag(s):
- [ ] No — reason:

### Checklist


src/main/services/ccdCase.interface.ts
```diff
@@ -167,6 +167,7 @@ export interface CcdDefendantParty {
   addressSameAsProperty?: string;
   phoneNumberProvided?: YesNoValue;
   phoneNumber?: string;
+  textMessageNumber?: string;
```
andrewdami11, line 172: This adds `textMessageNumber` on the frontend party type, but I could not find a matching field on `Party` / `PartyEntity` in `pcs-api` (only `phoneNumber` is there today). With unknown properties dropped on deserialize, mid-event draft save may not keep this value, so reload via `getInitialFormData` and any later texting would still have no number. May you provision the BE appropriately, as well?

src/main/steps/respond-to-claim/contact-preferences-text-message/index.ts
```diff
@@ -28,6 +28,35 @@ export const step: StepDefinition = createRespondToClaimFormStep({
         {
           value: 'yes',
           translationKey: 'options.yes',
+
+          subFields: {
+            mobileNumber: {
+              name: 'mobileNumber',
+              type: 'text',
+              required: true,
```
andrewdami11, line 58: The mobile subfield is required when Yes is selected, but `isAnswered` still only checks `defendantResponses.contactByText`. Someone who already answered Yes without a stored number could still look complete on the task list and skip past this step. Should Yes be treated as answered only when `textMessageNumber` is also present, similar to how telephone needs a number?

src/main/steps/respond-to-claim/contact-preferences-text-message/index.ts
```diff
@@ -58,10 +91,18 @@ export const step: StepDefinition = createRespondToClaimFormStep({
 
     if (contactByTextMessage === 'yes') {
       response.defendantResponses.contactByText = 'YES';
+      const mobileNumber = (req.body?.['contactByTextMessage.mobileNumber'] as string | undefined)?.trim();
```
andrewdami11: The validator strips spaces before checking (`07700 900 982` passes), but save only uses `trim()`, so the stored value can still contain spaces. Is it worth saving the same normalised digits-only value the validator uses, so what we store always matches what we accepted?

src/main/steps/respond-to-claim/contact-preferences-text-message/index.ts
```diff
@@ -58,10 +91,18 @@ export const step: StepDefinition = createRespondToClaimFormStep({
...
+      if (mobileNumber) {
+        response.defendantContactDetails.party.textMessageNumber = mobileNumber;
+      } else {
+        delete response.defendantContactDetails.party.textMessageNumber;
+      }
     } else if (contactByTextMessage === 'no') {
       response.defendantResponses.contactByText = 'NO';
+      delete response.defendantContactDetails.party.textMessageNumber;
```
andrewdami11, line 116: This step clears `textMessageNumber` when the text answer is No, which looks good. `normaliseContactPreferences` still only deletes `contactByText` when telephone becomes No, and does not clear `textMessageNumber` on the party. Is it worth clearing `textMessageNumber` there too, so a stale mobile is not left behind if the user later changes telephone to No?

src/test/unit/steps/respond-to-claim/contact-preferences-submit.test.ts
```diff
@@ -103,6 +104,11 @@ describe('contact preferences submit-time CCD payloads', () => {
       expect(saveDraftDefendantResponse).toHaveBeenCalledWith(
         req,
         expect.objectContaining({
+          defendantContactDetails: expect.objectContaining({
+            party: expect.objectContaining({
+              textMessageNumber: '07700900982',
```
andrewdami11, line 115: This covers the Yes path with a valid mobile, which is helpful. I could not see coverage for clearing `textMessageNumber` on No however, or for rejecting a landline or spaced-invalid value in the new validator. Would it be worth adding those cases?

src/main/services/ccdCase.interface.ts
ayub-macalim: the field now exists throughout the backend,  heres the PR for it - https://github.com/hmcts/pcs-api/pull/2398

src/main/steps/respond-to-claim/contact-preferences-text-message/index.ts
ayub-macalim: isAnswered now returns true for NO, but for YES it also requires the number

src/main/steps/respond-to-claim/contact-preferences-text-message/index.ts
ayub-macalim: Save now applies the same normalisation the validator uses

src/main/steps/respond-to-claim/contact-preferences-text-message/index.ts
ayub-macalim: It now also strips the party's mobile whenever the defendant isn't opted into text

src/test/unit/steps/respond-to-claim/contact-preferences-submit.test.ts
ayub-macalim: Addressed in contact-preferences-submit.test.ts

src/main/services/ccdCase.interface.ts
andrewdami11: Nicely implemented 👍🏽

May you add the correspondent label in this FE PR then (that links the BE PR), so the ticket can be comprehensively tested/validated?

Did you do that validation locally?

ayub-macalim: Yes tested locally it saves to draft case field

andrewdami11 (approved):

shazanm-soli (approved):
LGTM

PR #1858, opened 2026-08-06
Hdpi 6643 counter claim lr payment

m014389b:
### Jira link

See [PROJ-XXXXXX](https://tools.hmcts.net/jira/browse/PROJ-XXXXXX)

### Change description

image

image

- [ ] Yes
- [ ] No

### Feature flag assessment

**Feature flag:** Does this PR introduce or change user-facing behaviour that should be behind feature flag(s)?

- [ ] Yes — flag(s):
- [ ] No — reason:

### Checklist


src/main/assets/js/counter-claim-payment-choice.ts
```diff
@@ -0,0 +1,23 @@
+// pay btn stays outside the radio conditionals; card just GET-navigates
```
linusnorton: How does this work without JS? Maybe you should just detect the selected option in a GET handler on the server side then send the appropriate 301?

sadmanrahh: Good spot. I’ve added a server-side fallback for the no-JS case: when the legal-rep form is posted with the card option selected, `resolveRedirectAfterPost` now redirects to the existing card payment start route. The JS remains only as progressive enhancement so card users navigate directly, but the flow now works without JS as well. Added unit coverage for the card POST path.

src/main/routes/counterClaimPaymentStart.ts
```diff
@@ -79,4 +66,75 @@ export default function counterClaimPaymentStartRoutes(app: Application): void {
...
+      if (!accessToken) {
+        return redirectOnMissingAccessToken(caseReference, next);
+      }
+
+      if (!serviceRequestReference || feeAmount === undefined) {
+        return redirectOnMissingPaymentSessionData(res, caseReference);
+      }
+
+      try {
+        const paymentResponse = await paymentService.startPbaPaymentRequest({
```
linusnorton, line 91: Do we need to check the status of the response? like what if it fails?

src/main/steps/respond-to-claim/counter-claim-application-fee-amount/index.ts
```diff
@@ -1,33 +1,91 @@
...
+    }
+
+    const caseReference = req.params.caseReference;
+    const paymentOption = req.body?.paymentOptions as string | undefined;
+
+    if (paymentOption === 'card') {
+      return caseReference ? `/case/${caseReference}/respond-to-claim/counter-claim-payment/start` : '#';
+    }
+
+    if (paymentOption === 'pba') {
```
linusnorton: these need to be if else and we need an else to catch other things the user could put in

src/main/routes/counterClaimPaymentStart.ts
sadmanrahh: Yes. I’ve added an explicit status check now. We only persist the payment reference and go to the success page when the PBA response maps to a successful outcome; otherwise we redirect back to the fee page with the payment failed state. Added unit coverage for an unsuccessful PBA status.

src/main/steps/respond-to-claim/counter-claim-application-fee-amount/index.ts
sadmanrahh: Done. I’ve changed this to an explicit if/else-if/else and added a fallback for unexpected payment option values so a manipulated POST does not fall through to normal step navigation. Added unit coverage for that path.

src/main/steps/respond-to-claim/counter-claim-application-fee-amount/counterClaimApplicationFeeAmount.njk
```diff
@@ -32,9 +36,88 @@
...
+          } if fieldErrors['paymentOptions.pbaAccount']
+        }) }}
+
+      {{ govukInput({
+          id: 'customerReference',
+          name: 'paymentOptions.customerReference',
+          label: {
+            text: labels.customerReferenceHeading,
+            classes: "govuk-label--m",
+            isPageHeading: true
```
andrewdami11: This page already has an `<h1>` for the fee title, and `isPageHeading: true` on the customer reference label adds a second page heading when PBA is selected. Is accessibility going to _push back_ on this?

src/main/steps/respond-to-claim/counter-claim-application-fee-amount/index.ts
```diff
@@ -72,6 +140,53 @@ export const step: StepDefinition = createRespondToClaimFormStep({
...
+}
+
+async function getPbaAccounts(req: Request): Promise<string[]> {
+  const accessToken = req.session.user?.accessToken;
+  if (!accessToken) {
+    logger.error('Unable to get PBA accounts for user');
+    return [];
+  }
+
+  const pbaAccountsResponse = await paymentService.getPbaAccounts(accessToken);
```
andrewdami11: If `paymentService.getPbaAccounts` throws, the whole fee page GET can fail for legal reps instead of still showing the form. Is it worth catching that and falling back to an empty account list, similar to the _card_ payment failure pattern?

src/main/routes/counterClaimPaymentStart.ts
```diff
@@ -79,4 +66,87 @@ export default function counterClaimPaymentStartRoutes(app: Application): void {
...
```
andrewdami11, line 91: This route checks service request and fee amount, but not `pbaAccount` or `customerReference`, so a direct GET to the PBA start URL could call the payment API with missing values. Should this be _properly_ handled?

src/main/steps/respond-to-claim/counter-claim-application-fee-amount/index.ts
```diff
@@ -1,33 +1,91 @@
...
+            translationKey: {
+              label: 'labels.pba',
+            },
+            errorMessage: 'errors.pbaAccount',
+          },
+          customerReference: {
+            name: 'customerReference',
+            type: 'text',
+            required: true,
+            translationKey: { label: 'labels.customerReference' },
```
andrewdami11: The subField config uses `labels.customerReference`, but the locale files define `labels.customerReferenceHeading`. Is it worth aligning the key?

src/main/modules/steps/formBuilder/formFieldConfig.interface.ts
```diff
@@ -6,9 +6,17 @@ import type { DocumentStorage } from '@modules/documents/storage';
 import type { UploadValidationOptions } from '@utils/documentUploadValidation';
 
 export type FormFieldType =
-  'radio' | 'checkbox' | 'text' | 'date' | 'textarea' | 'character-count' | 'postcodeLookup' | 'file';
+  'radio' | 'checkbox' | 'text' | 'date' | 'textarea' | 'character-count' | 'postcodeLookup' | 'file' | 'select';
```
andrewdami11, line 9: `select` is added to the form field types here, but `componentBuilders` still falls through to a text `input` for unknown types.
Is it worth adding a proper `select` case in `componentBuilders` while the type is being introduced, so future form-builder selects do not render as text inputs?

src/main/steps/respond-to-claim/counter-claim-application-fee-amount/counterClaimApplicationFeeAmount.njk
sadmanrahh: Good point. I’ve removed `isPageHeading: true` from the customer reference label, so the page keeps the fee title as the only page heading.

src/main/steps/respond-to-claim/counter-claim-application-fee-amount/index.ts
sadmanrahh: Agreed. I’ve wrapped the PBA account lookup in a try/catch and fall back to an empty account list if the API call fails, so the fee page can still render. Added unit coverage for that fallback.

src/main/routes/counterClaimPaymentStart.ts
sadmanrahh: Yes, agreed. I’ve added a guard before the PBA API call so a direct GET without `pbaAccount` or `customerReference` redirects back to the fee page and does not call the payment API. Added unit coverage.

src/main/steps/respond-to-claim/counter-claim-application-fee-amount/index.ts
sadmanrahh: Done. I’ve aligned the subField config to use `labels.customerReferenceHeading`, which is the key defined in the locale files.

src/main/modules/steps/formBuilder/formFieldConfig.interface.ts
sadmanrahh: Agreed. I’ve added an explicit `select` case in `componentBuilders` and updated the schema component type list, with unit coverage for select rendering.

src/test/ui/data/page-data/adjustmentToGetIntoBuilding.page.data.ts
madhavi1203: Hi @m014389b, I noticed that a number of test files have been modified. As per our agreement, developers should not make changes to test files. Could you please revert all changes related to src/test/ui?

